Issue #16, opened 2020-07-16
Setup CI for macOS


nbaksalyar:
Use GitHub Actions or Travis to run continuous integration on macOS.

It's likely that code signing & permissions will be a bit of a problem.

List of resources:
- [lldb documentation on code signing](https://lldb.llvm.org/resources/build.html#code-signing-on-macos)
- [Apple's documentation on code signing](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Introduction/Introduction.html#//apple_ref/doc/uid/TP40005929-CH1-SW1)

nbaksalyar:
> just adding "- osx" as another OS in travis config file should work

That's good spotting, thanks!

The freezing on the integration test is semi-expected - it relies on reading memory of _another_ process as opposed to unit tests, which read memory of their own process, so this is likely caused by insufficient privileges of a parent process.

A short-term solution could be `#[ignore]`ing the integration test for macOS and enabling it for Linux (you can do that with `cargo test -- --ignored`).

If you want to submit a pull request that does that, it'd be very welcome!

JJendryka:
Should be closed by #24

PR #24, opened 2020-07-20
Add macos CI

JJendryka:
Implements #16 . For the time being disables read_memory integration test for macos. When testing on my fork Travis got stuck the first time it run, however after triggering build once again everything passed (I suspect a caching issue on Travis side). After that all tests worked as they should.


nbaksalyar:
Looks great, thank you!

PR #26, opened 2020-07-20
Improve OSX build instructions

fabiim:
The previous instructions did not work in my OSX. 
I had to fill in a full path to the Info.plist file for it to work. 
Not sure if this is the best option, but hey... it works.


nbaksalyar:
Good catch, thanks! I haven't found a simple way to provide a relative path either, so probably it will require some `build.rs` trickery in the future.

PR #28, opened 2020-07-23
Split out code common for all *nix platform

### LINKED ISSUE #9: Split out code common for all *nix from the Linux target module
FreeBSD, OpenBSD, NetBSD, and Illumos/Solaris all use `ptrace(2)` for process control & debuggers. These functions can be shared across `#[cfg(unix)]` OSes.

The main differences are in the way memory is read & written - e.g., FreeBSD supports [`PT_IO` request](https://www.freebsd.org/cgi/man.cgi?query=ptrace) type to write arbitrary amounts of data and Illumos mostly relies on an extended procfs for the same functionality.

DeltaManiac:
Hi @nbaksalyar 

### Changes 
- Moved the common code from linux to unix module
- Added `cirrus.yml` to test for BSD (`cargo build` and `cargo test`) on Cirrus CI and it [passes ](https://cirrus-ci.com/task/5952871088783360) for FreeBSD12

### Pending Tasks
- [x] Clean the commented code
- [x] Improve the doc Comments
- [ ] Add Cirrus badge to ReadMe.md

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -5,7 +5,8 @@ use std::{
     mem,
 };
 
-use crate::target::unix::UnixTarget;
+use crate::target::unix::{self, UnixTarget};
+
 /// This structure holds the state of a debuggee on Linux based systems
 /// You can use it to read & write debuggee's memory, pause it, set breakpoints, etc.
 pub struct LinuxTarget {
@@ -22,7 +23,7 @@ impl UnixTarget for LinuxTarget {
 impl LinuxTarget {
     /// Launches a new debuggee process
     pub fn launch(path: &str) -> Result<LinuxTarget, Box<dyn std::error::Error>> {
-        let pid = crate::target::unix::launch(path)?;
+        let pid = unix::launch(path)?;
         Ok(LinuxTarget { pid })
     }
 
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -1,20 +1,14 @@
-use nix::{
-    sys::ptrace,
-    sys::wait::waitpid,
-    unistd::{execv, fork, getpid, ForkResult, Pid},
-};
+use nix::unistd::{getpid, Pid};
 use std::{
-    ffi::CString,
     fs::File,
     io::{BufRead, BufReader},
     mem,
 };
 
+use crate::target::unix::Target;
+
 /// This structure holds the state of a debuggee.
 /// You can use it to read & write debuggee's memory, pause it, set breakpoints, etc.
-pub struct Target {
-    pid: Pid,
-}
 
 impl Target {
     /// Uses this process as a debuggee.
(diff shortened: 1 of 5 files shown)


tests/readmem.rs
```diff
@@ -33,7 +34,7 @@ fn read_memory() -> Result<(), Box<dyn std::error::Error>> {
 
     assert_eq!(&rval, b"Hello, world!");
 
-    target.unpause()?;
+    target.unpause(&target)?;
```
DeltaManiac, line 37: Hi @nbaksalyar 
I don't think this is an Idiomatic API by design, and its mostly cause of the weird structure in the `UnixTarget` trait.
Could you please provide your inputs here.

nbaksalyar: Trais in Rust don't have properties yet, but we can simply work around by having an abstract function like this:

```rust
trait UnixTarget {
  fn pid(&self) -> Pid;
}
```

Then it can be implemented e.g. for the Linux target struct:

```rust
impl LinuxTarget {
  fn pid(&self) -> Pid { self.pid }
}
```

and then in `unpause` it should suffice to call `self.pid()` to obtain the process ID.
And since all *nixes have the concept of PID, it's a good idea to introduce this getter function anyway!

src/target/unix.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+use std::ffi::CString;
+impl UnixTarget for Target {}
+pub struct Target {
+    pub pid: Pid,
+}
+
+pub trait UnixTarget {
+    /// Launches the debugee from the path provided
+    fn launch(path: &str) -> Result<Target, Box<dyn std::error::Error>> {
```
nbaksalyar, line 14: This probably can return just `Result<Pid, Error>` instead of `Result<Target, ...>`, and then it can be moved to a `pub(crate)` free function (i.e. not necessarily be a part of the trait). Then, in `LinuxTarget`, we can use it as follows: 

```rust
struct LinuxTarget { pid: Pid }

impl LinuxTarget {
  fn launch(&self, path: &str) -> Result<LinuxTarget, ...> {
    let pid = target::unix::launch(path);
    LinuxTarget { pid }
  }
}
```

And then, `UnixTarget::launch` (the trait function) can just define a generic interface without a blanket implementation, e.g.:

```rust
trait UnixTarget {
  fn launch(&self, path: &str) -> Result<Box<dyn UnixTarget>, ...>;
```

What do you think?

DeltaManiac: Thanks for the review. 😄 
I agree with the idea of making `launch` a free function, but then i couldn't quite grasp why we would need it in the `UnixTarget` too because the functionality of `launch` is wrapped in `LinuxTarget::launch`.
Am i missing something serious here, cause it feels like i am.
Could you please elaborate a bit more on this or guide me to some reading material ?

tests/readmem.rs
DeltaManiac: Done

src/target/unix.rs
nbaksalyar: >  Could you please elaborate a bit more on this

Sure! It boils down to the question of how do we create debug tools which can work on any supported target platform with a minimal number of changes required from a user. And you're right -- sorry for the confusion, `launch` should be a part of the concrete target type, actually. I was thinking in terms of using the `Target` trait as such:

```rust
trait Target {
  fn launch(path: &str) -> Box<dyn Target> where Self: Sized;
}

fn my_debug_tool<T: Target>() {
  let target = T::launch("/usr/bin/bash");
  target.read();
  // read target's memory etc.
}
```

but it doesn't make a lot of sense and it introduces unnecessary complexity. I think it's easier & better to instantiate the `Target` trait object before we call `my_debug_tool(target)`, i.e.:

```rust
fn my_debug_tool<T: Target>(target: T) {
  target.read();
  // read target's memory etc.
}

fn main() {
  let target = NativeTarget::launch("/usr/bin/bash");
  my_debug_tool(target);
}
```

So yeah, it's better to leave `fn launch(..)` out of traits altogether :+1:

DeltaManiac: Thanks for the explanation.
I too feel that the second way is a bit more organic at least for me

```rust
fn my_debug_tool<T: Target>(target: T) {
  target.read();
  // read target's memory etc.
}

fn main() {
  let target = NativeTarget::launch("/usr/bin/bash");
  my_debug_tool(target);
}
```
So then i think this PR can be merged, if the review is done?

Regarding CIBadge for freeBSD we would have to do [little bit more steps](https://cirrus-ci.org/guide/quick-start/) and then update readme.
I can create a new issue for CirrusCi integration if required.

src/target/linux.rs
```diff
@@ -1,63 +1,39 @@
...
+use nix::unistd::{getpid, Pid};
 use std::{
-    ffi::CString,
     fs::File,
     io::{BufRead, BufReader},
     mem,
 };
 
-/// This structure holds the state of a debuggee.
+use crate::target::unix::UnixTarget;
```
nbaksalyar, line 8: I think this can be replaced with

```
use crate::target::unix::{self, UnixTarget};
```

and then on line 25 we can use a more concise module name:

```
let pid = unix::launch(path)?;
```

What do you think?

DeltaManiac: Yeah, that makes it neat.
Is there any configuration for clippy that would capture such lints/best practices?

nbaksalyar: As far as I am aware, clippy doesn't catch that. It's more of a minor code style preference, I guess.

src/target/unix.rs
DeltaManiac: Created #31 to track CirrusCI

PR #29, opened 2020-07-23
Use a rust executable in tests

### LINKED ISSUE #10: Rewrite test binary in Rust
The [current test binary](https://github.com/headcrab-rs/headcrab/blob/116077cd5d67c2adf7398cd32d56e3a1fd673a94/tests/hello.c) is written in C. It's a temporary measure to simplify working with debug info & predictable variable addresses.

Rewriting the test in Rust will require either disabling [PIE](https://en.wikipedia.org/wiki/Position-independent_code) in compilation flags (`relocation-model=dynamic-no-pic`), or calculating relative addresses (this can be done by calling [`Target::get_addr_range`](https://github.com/headcrab-rs/headcrab/blob/116077cd5d67c2adf7398cd32d56e3a1fd673a94/src/target/linux.rs#L79-L90)) on Linux. MacOS has an undocumented feature to disable ASLR so likely relative addresses calculation won't be required (also see if there's a similar option for Linux).

fabiim:
Fixes #10.

- I decided to introduce another folder (tests/testees) to keep
programs under test separated. This simplifies the
Makefile because we don't need to filter out the test code files like
readmem.rs

- I'm using ```--dynamic-no-pic``` to disable PIE (as suggested in #10).
According to the docs it should be used for *relocatable external
references, non-relocatable code* [1]. This guarantees that
executable symbols addresses are absolute and can be used for
referencing the symbols at runtime. I'm not sure what it's meant
by *relocatable external references*, but I'm assuming it's relate
with library references, and not relevant for now.

[1] -  https://doc.rust-lang.org/rustc/codegen-options/index.html


tests/testees/Makefile
```diff
@@ -0,0 +1,13 @@
+CC       = rustc
+CC_FLAGS = -g -C relocation-model=dynamic-no-pic
```
nbaksalyar, line 2: This default option is likely to be reconsidered in the future, as we'll need to account for PIC anyway, but at this time we should leave this at `dynamic-no-pic` for the sake of simplicity.

PR #32, opened 2020-07-24
Make `ReadMemory::read` unsafe and document safety

bjorn3:


philipc:
Do we also need to document that this stores a mutable raw pointer, and the user must ensure the value is not used again until after `apply` is called? (Or even better would be using lifetimes to get rust to check this for us.)

bjorn3:
I thought the lifetimes were already such that `val` can't be accessed it is safe to do so, but it wasn't. The latest commit does ensure this.

PR #33, opened 2020-07-24
Improve readmem test

bjorn3:
* Build the testees from the test
* Ensure that the test doesn't depend on any particular current working dir

### DIFF
--- a/tests/readmem.rs
+++ b/tests/readmem.rs
@@ -1,18 +1,40 @@
 //! This is a simple test to read memory from a child process.
 
+use std::sync::Once;
+
 #[cfg(target_os = "linux")]
 use headcrab::{symbol::Dwarf, target::LinuxTarget, target::UnixTarget};
 
-static BIN_PATH: &str = "./tests/testees/hello";
+static TESTEES_BUILD: Once = Once::new();
+
+/// Ensure that all testees are built.
+fn ensure_testees() {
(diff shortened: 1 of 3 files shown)


tests/readmem.rs
```diff
@@ -1,18 +1,40 @@
...
+use std::sync::Once;
+
 #[cfg(target_os = "linux")]
 use headcrab::{symbol::Dwarf, target::LinuxTarget, target::UnixTarget};
 
-static BIN_PATH: &str = "./tests/testees/hello";
+static TESTEES_BUILD: Once = Once::new();
+
+/// Ensure that all testees are built.
+fn ensure_testees() {
```
nbaksalyar, line 11: Hmm. I think we can move this part to a build script in `build.rs`, it should also automatically ensure that it's called only once for all tests and it will track changes in source files too.

bjorn3: This is only necessary for tests. If I added it to the build script it would also run when building just the library and every time a testee changes the whole library would be rebuild. I could extract this function to a library for all tests if you want though.

nbaksalyar: Yeah, good point, there's no way to run `build.rs` for tests only. In the long run we probably should move integration tests (and other components) into separate crates anyway, which will give us a lot more flexibility, but I feel this is out of scope of this PR for now.

Issue #38, opened 2020-07-24
Get current register values for a thread on Linux


nbaksalyar:
There should be a function to get the current values of:
- the instruction pointer register (`RIP` on x86_64)
- the stack pointer register (`RSP` on x86_64)
- the stack frame pointer (`RBP` on x86_64)
- general purpose registers (platform-specific, e.g. `RAX`, `RBX`, `RCX` and others on x86_64)

bjorn3:
Working on this. Though currently only for the single-thread case.

nbaksalyar:
Thank you! Single-thread case should be good too :+1:

theIDinside:
Is this issue opened or closed?

bjorn3:
Looks like I implemented this in #40.

PR #40, opened 2020-07-24
Read regs for the main thread on Linux

bjorn3:
cc #38 (only the main thread for now)

### DIFF
--- a/tests/readregs.rs
+++ b/tests/readregs.rs
@@ -1,0 +1,59 @@
+//! This is a simple test to read registers from a child process.
+
+mod test_utils;
+
+#[cfg(target_os = "linux")]
+use headcrab::{target::LinuxTarget, target::UnixTarget};
+
+static BIN_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tests/testees/hello");
+
+#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+#[test]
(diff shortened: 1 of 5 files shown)


tests/readregs.rs
```diff
@@ -0,0 +1,52 @@
...
+    assert_eq!(regs.r10, 0);
+    assert_eq!(regs.r9, 0);
+    assert_eq!(regs.r8, 0);
+    assert_eq!(regs.rax, 0);
+    assert_eq!(regs.rcx, 0);
+    assert_eq!(regs.rdx, 0);
+    assert_eq!(regs.rsi, 0);
+    assert_eq!(regs.rdi, 0);
+    // sys_execve syscall: https://github.com/torvalds/linux/blob/f359287765c04711ff54fbd11645271d8e5ff763/arch/x86/entry/syscalls/syscall_64.tbl#L70
+    assert_eq!(regs.orig_rax, 59);
```
nbaksalyar, line 36: This can probably be made into a const (with the same comment above):
```suggestion
    pub const SYSCALL_EXECVE: usize = 59;
    assert_eq!(regs.orig_rax, SYSCALL_EXECVE);
```

bjorn3: I named it `X86_64_SYSCALL_EXECVE` as x86 uses a different number.

PR #41, opened 2020-07-24
Write memory

nbaksalyar:
- [ ] Write memory using `process_vm_write` on Linux.
- [ ] Write memory using `ptrace` on Linux.
- [ ] Write memory using `mach_vm_write` on macOS.
- [ ] Add tests for writing to write-protected pages.


src/target/linux/writemem.rs
```diff
@@ -0,0 +1,133 @@
+use nix::unistd::Pid;
+use std::{marker::PhantomData, mem};
+
+/// Allows to write data to different locations in debuggee's memory as a single operation.
+/// This implementation can select different strategies for different memory pages.
+pub struct WriteMemory<'a> {
+    pid: Pid,
+    write_ops: Vec<WriteOp>,
+    _marker: PhantomData<&'a mut ()>,
```
bjorn3, line 9:
```suggestion
    _marker: PhantomData<&'a ()>,
```

`WriteMemory` only requires immutable references, not mutable ones. This changes the variance of the lifetime parameter to allow a bit more code.

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,133 @@
...
+            pid,
+            write_ops: Vec::new(),
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn write<T>(mut self, val: &'a T, remote_base: usize) -> Self {
+        self.write_ops.push(WriteOp {
+            remote_base,
+            source_len: mem::size_of::<T>(),
```
bjorn3, line 24:
```suggestion
    pub fn write<T: ?Sized>(mut self, val: &'a T, remote_base: usize) -> Self {
        self.write_ops.push(WriteOp {
            remote_base,
            source_len: mem::size_of_val(val),
```

This allows passing in a slice to write.

src/target/linux/writemem.rs
nbaksalyar: Good catch, thanks!

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+/// Allows to write to write-protected pages.
+/// On Linux, this will result in multiple system calls and it's inefficient.
+pub(crate) fn write_ptrace(
+    _pid: Pid,
+    _write_ops: &[WriteOp],
+) -> Result<(), Box<dyn std::error::Error>> {
+    Ok(())
```
bjorn3, line 70: TODO

src/target/linux/mod.rs
```diff
@@ -38,6 +41,11 @@ impl LinuxTarget {
         ReadMemory::new(self.pid())
```
bjorn3, line 41: Should `ReadMemory` also be moved to a new file?

nbaksalyar: Definitely - will do in a next commit :)

src/target/linux/writemem.rs
```diff
@@ -44,35 +47,96 @@ pub(crate) struct WriteOp {
...
             iov_base: self.source_ptr as *mut _,
             iov_len: self.source_len,
         }
     }
+
+    /// Breaks the memory write operation into groups of words suitable for writing
+    /// with `ptrace::write`.
+    unsafe fn as_ptrace(&self) -> Vec<WriteOp> {
+        let mut output =
+            Vec::with_capacity(f64::ceil(self.source_len as f64 / WORD_SIZE as f64) as usize);
```
bjorn3, line 70:
```suggestion
        let mut output = Vec::with_capacity((self.source_len + WORD_SIZE - 1) / WORD_SIZE));
```

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.write_ops.push(WriteOp {
+            remote_base,
+            source_len: mem::size_of_val(val),
+            source_ptr: val as *const T as *const libc::c_void,
+        });
+        self
+    }
+
+    /// Executes the memory write operation.
+    pub unsafe fn apply(self) -> Result<(), Box<dyn std::error::Error>> {
```
bjorn3, line 33: This is safe as it can only crash the remote process, right?

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,301 @@
...
+        let write_mem = WriteMemory::new(getpid())
+            .write(&var, &write_var_op as *const _ as usize)
+            .write(&var2, &write_var2_op as *const _ as usize);
+
+        unsafe {
+            write_process_vm(write_mem.pid, &write_mem.write_ops).expect("Failed to write memory")
+        };
+
+        assert_eq!(write_var2_op, var2);
+        assert_eq!(write_var_op, var);
```
bjorn3, line 198: These should use `volatile_read`, otherwise LLVM assumes that nobody writes to them.

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,301 @@
...
+                        .expect("Failed to write memory")
+                };
+
+                ptrace::cont(parent, Some(signal::Signal::SIGCONT)).unwrap();
+            }
+            Ok(ForkResult::Parent { child, .. }) => {
+                wait::waitpid(child, None).unwrap();
+
+                assert_eq!(write_var_op, var);
+                assert_eq!(write_var2_op, var2);
```
bjorn3, line 232: Same here

src/target/linux/writemem.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/// A single memory write operation.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) struct WriteOp {
+    // Remote destation location.
+    remote_base: usize,
+    // Pointer to a source.
+    source_ptr: *const libc::c_void,
+    // Size of `source_ptr`.
```
bjorn3, line 45:
```suggestion
    /// Remote destation location.
    remote_base: usize,
    /// Pointer to a source.
    source_ptr: *const libc::c_void,
    /// Size of `source_ptr`.
```

These show up when using `--document-private-items`.

JJendryka:
I can help with adding test for writing to write-protected pages if you want?

nbaksalyar:
> I can help with adding test for writing to write-protected pages if you want?

No worries, I should finish this today. :) Thanks!

PR #43, opened 2020-07-24
Attach to process on linux

### LINKED ISSUE #39: Attach to running processes on Linux
There should be a function `LinuxTarget::attach(pid: Pid) -> Target` to start debugging an already existing process.
Use [`nix::ptrace::attach(pid)`](https://docs.rs/nix/0.17.0/nix/sys/ptrace/fn.attach.html) for this.

JJendryka:
Implements attaching to running process on linux. Closes #39. Cannot implement tests because most linux distros have capability to attach to different process disabled.

### DIFF
--- a/tests/attach_readmem.rs
+++ b/tests/attach_readmem.rs
@@ -11,9 +11,10 @@ mod test_utils;
 #[cfg(target_os = "linux")]
 use headcrab::{symbol::Dwarf, target::LinuxTarget, target::UnixTarget};
 
-static BIN_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tests/testees/hello");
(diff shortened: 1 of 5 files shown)


bjorn3:
Maybe add a `#[ignore]` test?

tests/attach_readmem.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let debuginfo = Dwarf::new(BIN_PATH)?;
+
+    let str_addr = debuginfo
+        .get_var_address("STATICVAR")
+        .expect("Expected static var has not been found in the target binary");
+
+    match fork()? {
+        ForkResult::Parent { child, .. } => {
+            let target = LinuxTarget::attach(child)?;
```
bjorn3, line 31: There is a race condition here. The child may already have exited when you try to attach. Maybe add another testee that has a busy loop and then wait like 10ms before attaching and at the end kill the testee.

PR #34, opened 2020-07-24
Refactor dwarf loading

bjorn3:
This stores the parsed object and DWARF together with the `Mmap` in `Dwarf` by using the rental crate for self referential structs. This is a prerequisite for lazily parsing debuginfo.

PR #46, opened 2020-07-25
Resolve symbols

### LINKED ISSUE #44: Map addr<->symbol name
I am working on this.

bjorn3:
Fixes #44
Based on #34

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,8 @@
 //! Headcrab, a modern Rust debugging library.
 
+#[macro_use]
+extern crate rental;
+
 /// Functions to work with target processes: reading & writing memory, process control functions, etc.
 pub mod target;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,5 @@
 //! Headcrab, a modern Rust debugging library.
 
-#[macro_use]
-extern crate rental;
-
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1,5 +1,8 @@
 //! Headcrab, a modern Rust debugging library.
 
+#[macro_use]
+extern crate rental;
```
nbaksalyar, line 4: This crate seems to be archived & not maintained anymore and the author recommends to explore other solutions: https://github.com/jpernst/rental

As it seems, it's only used for safe borrowing of

bjorn3: I didn't know it was archived.

bjorn3: I have extracted the actual debuginfo parsing into `ParsedDwarf::new` keeping `Dwarf::new` as a thin wrapper that only handles the self-referential part. I have also written some unsafe code as replacement for rental. I am currently waiting on a review over at the `t-lang/wg-unsafe-code-guidelines` stream of the rust-lang zulip.

nbaksalyar: Thank you!

> some unsafe code as replacement for rental

This can be submitted as a separate issue/PR, if it would be easier for you.
I'm going to review the rest of the code now.

bjorn3: I got a review, so I pushed it.

PR #47, opened 2020-07-26
Get a list of threads from Linux debuggee

### LINKED ISSUE #36: Get a list of process threads on Linux
`Target::threads() -> &[Thread]` or a similar function should return a view into debuggee process's threads.
For this issue, it should suffice to return a list of thread IDs and thread names (if they're available), so `Thread` can have an API like this:

```rust
trait Thread {
  type ThreadId = usize;

  fn name(&self) -> Option<String>;
  fn thread_id(&self) -> Self::ThreadId;
}
```

Later, more functions can be added to `Thread`.

fabiim:
Fixes #36 

 - Add a Thread trait
 - Retrieves a list of threads in linux

Notes:

I dropped  a couple of the suggestions in the ticket (#36) because: 

- I decided to use procfs because libproc-rs does not support getting the list of threads from linux [right now](https://github.com/andrewdavidmackenzie/libproc-rs/blob/123d1762727c9c5d4e81888d8ea1a7f9701ddb45/src/libproc/proc_pid.rs#L260). I imagine this crate being useful in the future anyhow.

- Having a default associated type in the Threat trait is not stable. 

- Returning a slice of Threat references does not make much sense to me because the threats change over time and I feel it's more useful to get a fresh current view every time.


src/target/unix.rs
```diff
@@ -15,6 +16,10 @@ pub trait UnixTarget {
         ptrace::cont(self.pid(), None)?;
         Ok(())
     }
+
+    /// Returns the current snapshot view of this debugee process threads
+    fn threads(&self)
```
JJendryka, line 21: Why add threads function declaration to UnixTarget if no code is shared between unix targets? Shouldn't `fn threads()` be specific to Linux target just like read()?

.gitignore
```diff
@@ -4,3 +4,4 @@
 /target
 Cargo.lock
 .vscode/
+.idea/
```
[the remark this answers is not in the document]
fabiim: Fixed.

Cargo.toml
```diff
@@ -16,6 +16,9 @@ libproc = "0.7.2"
 libc = "0.2.72"
 object = "0.20"
 
+[target.'cfg(target_os="linux")'.dependencies]
+procfs = "0.8.0"
```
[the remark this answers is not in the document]
fabiim: Moved.

src/target/unix.rs
fabiim: I imaged all UnixTarget would have threads, and that you might possibly want it there... lol. 
I'll stop imagining things :smile:  ... Removed the code and added it to the LinuxTarget.

src/target/linux.rs
```diff
@@ -48,6 +75,24 @@ impl LinuxTarget {
     pub fn read_regs(&self) -> Result<libc::user_regs_struct, Box<dyn std::error::Error>> {
         nix::sys::ptrace::getregs(self.pid()).map_err(|err| err.into())
     }
+
+    /// Returns the current snapshot view of this debugee process threads
```
bjorn3, line 79:
```suggestion
    /// Returns the current snapshot view of this debugee process threads.
```

src/target/linux.rs
```diff
@@ -268,4 +317,52 @@ mod tests {
...
+
+        let proc = LinuxTarget::me();
+        let threads = proc.threads()?;
+
+        let threads: Vec<(String, usize)> = threads
+            .iter()
+            .map(|t| (t.name().unwrap().clone(), t.thread_id()))
+            .collect();
+
+        // Find  at least 3 threads (cargo runs a minimum of 2)
```
bjorn3, line 341:
```suggestion
        // Find at least 3 threads (cargo runs a minimum of 2)
```

src/target/linux.rs
```diff
@@ -268,4 +317,52 @@ mod tests {
...
+            .name("thread-name".to_string())
+            .spawn(move || {
+                t1_barrier.wait();
+            })
+            .unwrap();
+
+        let proc = LinuxTarget::me();
+        let threads = proc.threads()?;
+
+        let threads: Vec<(String, usize)> = threads
```
fabiim, line 336: I need to make sure the thread exists before calling this.

src/target/linux.rs
```diff
@@ -268,4 +317,52 @@ mod tests {
...
+            threads.iter().any(|&(_, tid)| tid == proc_pid),
+            "Expected to find main pid={} in {:?}",
+            proc_pid,
+            threads
+        );
+
+        // Find thread name
+        assert!(
+            threads.iter().any(|(name, _)| name == "thread-name"),
+            "Expected to find main pid={} in {:?}",
```
bjorn3, line 360:
```suggestion
            "Expected to find thread "thread-name" in {:?}",
```

src/target/linux.rs
bjorn3: Maybe use another barrier?

fabiim: fixed

src/target/linux.rs
```diff
@@ -48,6 +75,25 @@ impl LinuxTarget {
     pub fn read_regs(&self) -> Result<libc::user_regs_struct, Box<dyn std::error::Error>> {
         nix::sys::ptrace::getregs(self.pid()).map_err(|err| err.into())
     }
+
+    /// Returns the current snapshot view of this debugee process threads.
+    #[allow(dead_code)]
+    fn threads(
```
bjorn3, line 81: *pub fn

src/target/linux.rs
```diff
@@ -1,12 +1,39 @@
...
+    fn new(name: &str, id: usize) -> LinuxThread {
+        LinuxThread {
+            name: name.to_string(),
+            id,
+        }
+    }
+}
+
+impl Thread for LinuxThread {
+    type ThreadId = usize;
```
bjorn3, line 27: usize doesn't make sense for this. Please use whatevee type `task.tid` is.

src/target/linux.rs
fabiim: :facepalm:

src/target/linux.rs
fabiim: fixed

src/target/linux.rs
```diff
@@ -1,12 +1,39 @@
...
 };
 
-use crate::target::unix::{self, UnixTarget};
+struct LinuxThread {
+    name: String,
+    id: i32,
+}
+
+impl LinuxThread {
+    fn new(name: &str, id: i32) -> LinuxThread {
```
nbaksalyar, line 18: A minor suggestion: we can take `Into<String>` here to make the API more flexible:
```suggestion
    fn new(name: impl Into<String>, id: i32) -> LinuxThread {
        LinuxThread {
            name: name.into(),
            id,
        }
    }
```

src/target/linux.rs
```diff
@@ -268,4 +316,60 @@ mod tests {
...
+
+        let proc = LinuxTarget::me();
+        let threads = proc.threads()?;
+
+        let threads: Vec<_> = threads
+            .iter()
+            .map(|t| (t.name().unwrap().clone(), t.thread_id()))
+            .collect();
+
+        // Find at least 3 threads (cargo runs a minimum of 2)
```
nbaksalyar, line 347: > cargo runs a minimum of 2

This is not always the case and it's configurable using the `test-threads` parameter (e.g.: `cargo test -- --test-threads=1`) or the environment variable `RUST_TEST_THREADS`.

As it seems, currently there's no way to get a number of test threads from _inside_ the tests (reading the env var is inconsistent), but I guess for the moment we can do something like

```suggestion
        // Find at least 3 threads (cargo runs a minimum of 2)
        let cargo_threads = std::env::var("RUST_TEST_THREADS")
            .map(|s| s.parse::<usize>())
            .unwrap_or(2);
```

src/target/thread.rs
```diff
@@ -0,0 +1,6 @@
+pub trait Thread {
+    type ThreadId;
+
+    fn name(&self) -> Option<String>;
+    fn thread_id(&self) -> Self::ThreadId;
+}
```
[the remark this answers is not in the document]
nbaksalyar: Agreed, it looks like `Thread` should be almost like a subset of `Target`.
What are your thoughts on this @fabiim, @bjorn3?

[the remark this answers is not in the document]
bjorn3: We could say that `Target` has one or more `Thread`'s with one being the main thread. `read_regs` will then ve moved to `Thread`, but `read` and `write` remain part of `Target`. The `unpause` method should remain part of `Target` I think, at least for as long as we don't have an equivalent to gdb's non-stop mode in which individual threads are paused and continued.

nbaksalyar:
> currently there's no way to get a number of test threads from inside the tests (reading the env var is inconsistent)

FYI, I've raised a new issue for Rust: https://github.com/rust-lang/rust/issues/74845

src/target/linux.rs
fabiim: When I considered cargo 2 threads it was the main thread + the test thread. 
Still your comment applies, I will do that +1 :)

fabiim: [Reference](https://stackoverflow.com/questions/43458194/is-there-any-way-to-tell-cargo-to-run-its-tests-on-the-main-thread)

src/target/linux.rs
```diff
@@ -1,12 +1,44 @@
...
+    fn new(name: impl Into<String>, id: i32) -> LinuxThread {
+        LinuxThread {
+            name: name.into(),
+            id,
+        }
+    }
+        LinuxThread {
+            name: name.to_string(),
+            id,
+        }
```
bjorn3, line 27:
```suggestion
```

src/target/linux.rs
fabiim: should be fixed now

src/target/thread.rs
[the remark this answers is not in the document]
fabiim: Let me see if I get it: 

- Move the read/write methods to the thread trait 
- Add a way to retrieve a main thread from a target 

 If that is correct, I have a couple of questions: 

1. What should be the abstraction for the registers structure be in the trait? 
2. Should ```threads()``` return all threads (including the main thread?)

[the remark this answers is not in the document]
bjorn3: > What should be the abstraction for the registers structure be in the trait?

Just `type Registers: Debug;` or something like that for now I think. It is possible to later add additional bounds as necessary. I think `threads` should return a concrete thread type though for that to work.

> Should threads() return all threads (including the main thread?)

I don't have an opinion on this.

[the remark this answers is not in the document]
nbaksalyar: > Should threads() return all threads (including the main thread?)

Yes, I believe so. At least, most (all?) major debuggers impose this behaviour - e.g., if you launch a single-threaded process in gdb and run the `info threads` command, the output will include the main thread.

bjorn3:
```
target::linux::tests::reads_threads stdout ----

Error: NotFound(Some("/proc/5167/task/5170/stat"))
```

fabiim:
> ```
> target::linux::tests::reads_threads stdout ----
> 
> Error: NotFound(Some("/proc/5167/task/5170/stat"))
> ```

Yep, there is a bug, it should tolerate tasks that have gone away . I imagine this hit that. 
Fixing it...

src/target/linux.rs
```diff
@@ -137,6 +165,40 @@ impl LinuxTarget {
...
+        &self,
+    ) -> Result<Vec<Box<dyn Thread<ThreadId = i32>>>, Box<dyn std::error::Error>> {
+        let tasks: Vec<_> = Process::new(self.pid.as_raw())?
+            .tasks()?
+            .flatten()
+            .collect();
+
+        let mut result: Vec<Box<dyn Thread<ThreadId = i32>>> = vec![];
+        for task in tasks {
+            // The task page might be incomplete and we should retry
```
fabiim, line 180: This retry logic might be too overkill? I'm not even sure that we can get Incomplete for this particular case, just handling the generic error type (couldn't find a quick answer if it can happen on every page or just particular cases).

It's not really consistent with the other methods into procfs like tasks above... 

So what do you all thing I should do here? @nbaksalyar  @bjorn3

fabiim: I'm taking the retry out. If the lib user doesn't get a thread because it is incomplete, he just needs to call threads again.

fabiim:
> > ```
> > target::linux::tests::reads_threads stdout ----
> > 
> > Error: NotFound(Some("/proc/5167/task/5170/stat"))
> > ```
> 
> Yep, there is a bug, it should tolerate tasks that have gone away . I imagine this hit that.
> Fixing it...

Fixed. Now we just skip threads not found  and/or not ready yet.

src/target/linux.rs
```diff
@@ -137,6 +165,31 @@ impl LinuxTarget {
...
+            match task.stat() {
+                Ok(t_stat) => {
+                    let thread = LinuxThread::new(&t_stat.comm, task.tid);
+                    result.push(Box::new(thread));
+                }
+                Err(ProcError::NotFound(_)) | Err(ProcError::Incomplete(_)) => {
+                    // ok to skip. Thread is gone or it's page is not complete yet.
+                }
+                Err(err) => return Err(Box::new(err)),
+            }
```
bjorn3, line 189: A thread could change it's own name I think. Maybe LinuxThread should only store pid and tid and name() would read the current name or return None when the thread doesn't exist anymore?

fabiim: Ah... interesting. Ok will check this later today then. 

Regarding keeping the pid and tid: as far as I can tell they are the same thing.
So we only need to keep one. Is this correct?

bjorn3: `procfs::process::Task` contains the pid of the whole process and the tid of the specific thread. You could also directly store that struct.

fabiim: You are right, a thread can change the name. From [ man procfs](https://man7.org/linux/man-pages//man5/procfs.5.html): 
> A thread may modify its comm value, or that of any of other thread in the same thread group 

Going with your idea of having the ```name``` method return the most recent: do you think it will be fine to swallow all other [procfs::ProcError](https://docs.rs/procfs/0.8.0/procfs/enum.ProcError.html) into a None? If so, the user wouldn't know something went wrong. Should we use a result on the name? 

Also I'm not sure I understand the benefits of the change. If you are getting a list of threads you either have them all paused or as soon as you get your information it might already be outdated. So the user will always have this issue of seeing old data, even if we push it elsewhere. He will always need to poll the thread list to get up-to-date information.

bjorn3: > Should we use a result on the name?

Yes

> If you are getting a list of threads you either have them all paused or as soon as you get your information it might already be outdated.

I think once we support actually debugging threads, we would store a list of all active threads. When using `PTRACE_O_TRACECLONE` to automatically attach to new threads we have to manually continue the new thread anyway, so we could just as well store it in a list. Only when attaching to a new process do we need to list all active threads. Making the `LinuxThread` persistent also has the benefit that you could continue and pause individual threads rather than the whole process.

fabiim: Done. 

> I think once we support actually debugging threads, we would store a list of all active threads. When using PTRACE_O_TRACECLONE to automatically attach to new threads we have to manually continue the new thread anyway, so we could just as well store it in a list. Only when attaching to a new process do we need to list all active threads. Making the LinuxThread persistent also has the benefit that you could continue and pause individual threads rather than the whole process.

Alright, I completely missed that the first time you mentioned. I think I get it now - thank you!. We are going to be in lockstep with threads getting notified through ptrace that a new thread was created and adding that to the list of the process.

fabiim:
@nbaksalyar @bjorn3 Is this ok?

My understanding from the discussion is that 2 essential features are missing:

- The Thread abstraction should have (to start with) the read registers method using a Register: Debug Trait 
- The thread list should be kept inside the process and we need to update it as threads come and go 

But I think it might be better to handle them as separate PR's?

PR #49, opened 2020-07-26
Readmem tests

### LINKED ISSUE #42: Add tests for read-protected VM pages
- Implement a test to read memory from a read-protected location (use [`nix::sys::mman::mprotect`](https://docs.rs/nix/0.17.0/nix/sys/mman/fn.mprotect.html)).
- Test for cross-page boundary reads (example case: part of large struct / string is stored in a read-protected page).
- Expected results: `SIGSEGV` for `process_vm_read`, success for `ptrace(PTRACE_PEEKDATA)`.

JJendryka:
Closes #42.

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -169,6 +169,11 @@ mod tests {
     use super::ReadMemory;
     use nix::unistd::getpid;
 
+    extern crate alloc;
+    use alloc::alloc::{alloc_zeroed, Layout};
+
+    use nix::sys::mman::{mprotect, ProtFlags};
+
     #[test]
     fn read_memory() {
         let var: usize = 52;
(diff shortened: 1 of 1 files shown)


src/target/linux.rs
```diff
@@ -169,6 +169,11 @@ mod tests {
     use super::ReadMemory;
     use nix::unistd::getpid;
 
+    extern crate alloc;
+    use alloc::alloc::{alloc_zeroed, Layout};
```
nbaksalyar, line 173: I believe this can be replaced with
```suggestion
    use std::alloc::{alloc_zeroed, Layout};
```
and then we won't need `extern crate alloc`

src/target/linux.rs
```diff
@@ -188,4 +193,60 @@ mod tests {
...
+            let res = ReadMemory::new(getpid())
+                .read(&mut read_var_op, ptr as *const _ as usize)
+                .apply();
+
+            // Expected to fail when reading read-protected memory.
+            // FIXME: Change when reading read-protected memory is handled properly
+            match res {
+                Ok(()) => panic!("Unexpected result: reading protected memory succeeded"),
+                Err(_) => (),
+            }
```
nbaksalyar, line 224: `ptr` leaks memory here. We can deallocate it as follows:

```suggestion
            }
            mprotect(
                ptr as *mut std::ffi::c_void,
                PAGE_SIZE,
                ProtFlags::PROT_WRITE | ProtFlags::PROT_READ,
            )
            .expect("Failed to mprotect");
            std::alloc::dealloc(ptr, Layout::from_size_align(PAGE_SIZE, PAGE_SIZE).unwrap());
```

src/target/linux.rs
```diff
@@ -188,4 +193,60 @@ mod tests {
...
+
+            ReadMemory::new(getpid())
+                .read(&mut read_var_op, array_ptr as *const _ as usize)
+                .apply()
+                .expect("Failed to apply memop");
+
+            // Expected result because of cross page read
+            // FIXME: Change when cross page read is handled correctly
+            assert_eq!([123, 0], read_var_op);
+        }
```
nbaksalyar, line 250: Same thing about `ptr` here, it should be deallocated

Issue #50, opened 2020-07-26
Correct behavior when reading memory on Linux


JJendryka:
Right now memory on linux is read using `process_vm_readv`, however it EFAULT's when it is trying to read from read protected page and has incorrect partial read behavior. (If read crosses page boundary and one of the pages is read restricted then read returns 0 in place of restricted data). 
To read restricted memory PTRACE_PEEKTEXT can be used.
To avoid incorrect partial reads those should be split into more on page boundaries and after read is performed data should be merged back together.

JJendryka:
I think I will give it a go.

PR #52, opened 2020-07-27
Correct behavior for reading memory on linux

JJendryka:
Implements #50 

- [x] Use ptrace as a fallback for process_vm_readv
- [x] Handle cross page read correctly
- [x] Fix tests so that ptrace can be used inside them
- [x] Restructure the code to be more readable


src/target/linux.rs
```diff
@@ -144,13 +200,63 @@ impl<'a> ReadMemory<'a> {
             )
         };
 
-        if bytes_read == -1 {
-            // fixme: return a proper error type
-            return Err(Box::new(nix::Error::last()));
+
+        if bytes_read < read_len as isize {
+            let maps = Process::new(self.pid.as_raw())?.maps()?;
```
bjorn3, line 205: FYI #59 will add a `memory_maps` method. The returned `MemoryMap` type doesn't have a permission field, but it is easy to add.

src/target/linux.rs
```diff
@@ -169,20 +177,63 @@ impl ReadOp {
...
+                });
+                left -= page_size;
+            }
+        }
+        out
+    }
 }
 
 /// Allows to read memory from different locations in debuggee's memory as a single operation.
-/// On Linux, this will correspond to a single system call / context switch.
```
bjorn3, line 175:
```suggestion
/// On Linux, this will most of the time correspond to a single context switch.
```

src/target/linux.rs
```diff
@@ -201,34 +252,66 @@ impl<'a> ReadMemory<'a> {
...
+            let (protected, readable) = self.split_protected(&self.read_ops)?;
+
+            self.read_process_vm(&readable)?;
+            self.read_ptrace(&protected)?;
+        }
+        Ok(())
+    }
+
+    /// Allows to read from several different locations with one system call.
+    /// It will ignore pages that are not readable. Returns number of bytes read at granularity of ReadOps.
```
bjorn3, line 299:
```suggestion
    /// It will error on pages that are not readable. Returns number of bytes read at granularity of ReadOps.
```

src/target/linux.rs
```diff
@@ -238,12 +321,72 @@ impl<'a> ReadMemory<'a> {
...
+        let protected_maps = maps
+            .iter()
+            .filter(|map| !map.is_readable)
+            .collect::<Vec<_>>();
+
+        let (protected, readable): (_, Vec<_>) = read_ops.iter().partition(|read_op| {
+            protected_maps.iter().any(|map| {
+                map.address.0 as usize <= read_op.remote_base
+                    && read_op.remote_base < map.address.1 as usize
+            })
```
bjorn3, line 346: A binary search would be faster.

src/target/linux.rs
```diff
@@ -238,12 +321,72 @@ impl<'a> ReadMemory<'a> {
...
+        });
+
+        Ok((protected, readable))
+    }
 
+    /// Allows to read from protected memory pages.
+    /// This operation results in multiple system calls and is inefficient.
```
bjorn3, line 353:
```suggestion
    /// This operation results in multiple system calls and as a result is inefficient.
```

JJendryka:
> As far as I know `/proc/<pid>/maps` is always sorted by address.

I've also thought so, however I cannot find any resource that states so.

src/target/linux.rs
```diff
@@ -238,12 +321,84 @@ impl<'a> ReadMemory<'a> {
...
+
+        let maps = self.target.memory_maps()?;
+
+        let mut protected_maps = maps
+            .iter()
+            .filter(|map| !map.is_readable)
+            .collect::<Vec<_>>();
+
+        // Not sure if required
+        protected_maps.sort_by(|a, b| a.address.0.cmp(&b.address.0));
```
nbaksalyar, line 345: > I cannot find any resource that states so.

Yeah, it wasn't easy to find an authoritative source about this, but I think the Linux's [proc filesystem documentation](https://github.com/torvalds/linux/blob/master/Documentation/filesystems/proc.rst) can be considered one and the following statement can be interpreted as a guarantee that addresses always read in the growing order:

> Note: reading /proc/PID/maps or /proc/PID/smaps is inherently racy (consistent output can be achieved only in the single read call).
>
> This typically manifests when doing partial reads of these files while the memory map is being modified. Despite the races, we do provide the following guarantees:
>    1. The mapped addresses never go backwards, which implies no two regions will ever overlap.
>    2. If there is something at a given vaddr during the entirety of the life of the smaps/maps walk, there will be some output for it.

So I believe it should be safe to remove this line.

PR #54, opened 2020-07-27
Add disassembler and line info parsing

bjorn3:
Blocked on https://github.com/gimli-rs/addr2line/issues/187 to get addr2line working files compiled from assembly. Otherwise the line info test fails.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ readme = "README.md"
 lazy_static = "1.4.0"
 object = "0.20"
 gimli = "0.22.0"
+capstone = "0.7.0"
 
 # Dependencies specific to macOS & Linux
 [target.'cfg(unix)'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ lazy_static = "1.4.0"
 object = "0.20"
 gimli = "0.22.0"
 capstone = "0.7.0"
+addr2line = "0.13.0"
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -15,6 +15,8 @@ nix = "0.17.0"
 libproc = "0.7.2"
 libc = "0.2.72"
 object = "0.20"
+capstone = "0.7.0"
```
nbaksalyar, line 18: Just a general comment: one alternative to consider is [iced-x86](https://github.com/0xd4d/iced/tree/master/src/rust/iced-x86) which is written in pure Rust (though it lacks support for ARM and other ISAs at this time). It's not important at this stage, but I think in the longer term targeting 100% Rust code might be better.

bjorn3: > iced-x86 is a high performance and **correct** x86 (16/32/64-bit) instruction decoder, disassembler and assembler written in Rust.

Being "correct" is a bold claim for an x86 disassembler. x86 is extremely complex.

nbaksalyar: Yeah, I see that Capstone is also used by Radare2 and Frida, I was unaware of that.
A library with such a wide community support is definitely preferable!

bjorn3:
I switched the `source.rs` test to use `hello` instead of `known_asm`. This means that gimli-rs/addr2line#187 is no longer a problem. (There is already a fix for it on master by the way, but is is not yet released)

bjorn3:
You noticed the branch names! :)

PR #57, opened 2020-07-27
Add a simple repl example

bjorn3:
It only supports a subset of the functionality of headcrab, but it is a good example of how to use headcrab.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,6 @@ memmap = "0.7.0"
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 mach = "0.3"
 security-framework-sys = "1.0"
+
+[dev-dependencies]
+rustyline = "6.2.0"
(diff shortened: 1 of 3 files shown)


nbaksalyar:
This is very exciting, thanks! :)
If it's easier, the example can be disabled for macOS temporarily - we can create a separate issue to address it.

bjorn3:
I was already looking for a way to disable it on macOS. I can't find any.

nbaksalyar:
>  I can't find any.

One hack is to use an empty `fn main` for all targets other than Linux and put all other code into an internal module.

E.g.:

```rust
#[cfg(target_os = "linux")]
fn main() {
  example::main();
}

#[cfg(not(target_os = "linux"))]
fn main() {
  println!("This example is not supported for OSes other than Linux");
}
```

and then the example implementation can be hidden in an internal module in a similar way:

```
#[cfg(target_os = "linux")]
mod example {
  struct Context { ... }

  fn main() {
    // the actual main function
  }
}
```

PR #58, opened 2020-07-28
More execution commands and use waitpid where sensible

bjorn3:
While there are legitimate use cases where waitpid should not be used, not using it will in most cases lead to race-conditions between stopping execution of the debuggee again and reading/writing state.

### DIFF
--- a/src/target/unix.rs
+++ b/src/target/unix.rs
@@ -24,7 +24,7 @@ pub trait UnixTarget {
         Ok(status)
     }
 
-    /// Detach from the debuggee, continuing it's execution.
+    /// Detach from the debuggee, continuing its execution.
     fn detach(&self) -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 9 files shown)


src/target/unix.rs
```diff
@@ -10,25 +10,45 @@ pub trait UnixTarget {
...
     /// Continues execution of a debuggee.
-    fn unpause(&self) -> Result<(), Box<dyn std::error::Error>> {
+    fn unpause(&self) -> Result<WaitStatus, Box<dyn std::error::Error>> {
         ptrace::cont(self.pid(), None)?;
-        waitpid(self.pid(), None).unwrap();
+        let status = waitpid(self.pid(), None)?;
+        Ok(status)
+    }
+
+    /// Detach from the debuggee, continuing it's execution.
```
nbaksalyar, line 27: Nit:
```suggestion
    /// Detach from the debuggee, continuing its execution.
```

PR #59, opened 2020-07-28
Handle pie executables on Linux

### LINKED ISSUE #51: Handle PIE executables on Linux by reading mem maps

bjorn3:
Fixes #51

### DIFF
--- a/src/symbol/mod.rs
+++ b/src/symbol/mod.rs
@@ -3,13 +3,14 @@
 
 use gimli::read::{EvaluationResult, Reader as _};
 use object::{
-    read::{Object, ObjectSection, Symbol},
+    read::{Object, ObjectSection, ObjectSegment, Symbol},
     SymbolKind,
 };
 use std::{
     borrow::Cow,
     collections::{BTreeMap, HashMap},
     fs::File,
+    path::Path,
     rc::Rc,
 };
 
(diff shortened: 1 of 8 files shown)


src/target/unix.rs
```diff
@@ -32,6 +33,12 @@ pub(crate) fn launch(path: &str) -> Result<Pid, Box<dyn std::error::Error>> {
         ForkResult::Child => {
             ptrace::traceme()?;
 
+            // Disable ASLR
+            #[cfg(target_os = "linux")]
+            unsafe {
+                libc::personality(0x0040000 /*ADDR_NO_RANDOMIZE*/);
```
nbaksalyar, line 39: Nit: this probably would be better as
```suggestion
               const ADDR_NO_RANDOMIZE: usize = 0x0040000;
               libc::personality(ADDR_NO_RANDOMIZE);
```
It also seems that this const is missing in rust-lang/libc, I'm going to raise a PR for that :)

src/target/mod.rs
```diff
@@ -12,3 +12,11 @@ pub use linux::*;
 mod macos;
 #[cfg(target_os = "macos")]
 pub use macos::*;
+
+#[derive(Debug)]
+pub struct MemoryMap {
```
nbaksalyar, line 17: Nice one, this is also helpful for the `macos` target :+1:

src/symbol/mod.rs
```diff
@@ -277,3 +280,132 @@ impl Dwarf {
         self.rent(|parsed| parsed.get_var_address(name))
     }
 }
+
+pub struct RelocatedDwarf(Vec<RelocatedDwarfEntry>);
```
nbaksalyar, line 284: This struct probably can be moved into its own module, but can be done later.

src/target/unix.rs
bjorn3: Yeah, all personalities are missing. I had to `cat /usr/include/sys/personality.h` to find the correct value.

src/symbol/mod.rs
bjorn3: Will do this later.

nbaksalyar:
> It also seems that this const is missing in rust-lang/libc, I'm going to raise a PR for that :)

Raised https://github.com/rust-lang/libc/pull/1849

PR #62, opened 2020-07-30
Kill debugee on debugger exit

### LINKED ISSUE #48: Use PTRACE_O_EXITKILL on Linux
This prevents the tracee from staying alive in a suspended state when the debugger exits.

JJendryka:
Closes #48. If debugee was launched by debugger it is killed when debugger exits.

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -27,7 +27,9 @@ impl LinuxTarget {
         path: &str,
     ) -> Result<(LinuxTarget, nix::sys::wait::WaitStatus), Box<dyn std::error::Error>> {
         let (pid, status) = unix::launch(path)?;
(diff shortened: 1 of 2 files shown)


src/target/linux.rs
```diff
@@ -141,6 +143,12 @@ impl LinuxTarget {
             })
             .collect())
     }
+
+    /// Kill debuggee when debugger exits.
+    fn kill_on_exit(&self) -> Result<(), Box<dyn std::error::Error>> {
```
nbaksalyar, line 148: I think this can be a `pub fn` for the case when a user wants to use this option after `attach`, but it's unclear whether this should be a function call, or a configuration parameter like `target.attach(pid, Config { kill_on_exit: true, ... })`. Let's leave it as it is now & create an issue for future reference.

PR #64, opened 2020-07-30
Add attach options

### LINKED ISSUE #63: Allow a user to decide if they want to kill the target after detach
As was discussed in #48, a user should have an option to automatically kill the debuggee after a call to `target.attach(pid)`. PR #62 implemented a new function, `kill_on_exit()` which sets a corresponding option. We should either make this function public, or add a new parameter for `Target::attach` to specify options (e.g.: `Target::attach(pid, AttachOptions { kill_on_exit: true })`).

JJendryka:
Closes #63

### DIFF
--- a/examples/repl.rs
+++ b/examples/repl.rs
@@ -138,8 +138,12 @@ mod example {
                 if let Some(pid) = parts.next() {
                     let pid = nix::unistd::Pid::from_raw(pid.parse()?);
                     println!("Attaching to process {}", pid);
-                    let (remote, status) =
-                        LinuxTarget::attach(pid, AttachOptions { kill_on_exit: true })?;
+                    let (remote, status) = LinuxTarget::attach(
+                        pid,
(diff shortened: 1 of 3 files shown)


examples/repl.rs
```diff
@@ -138,7 +138,8 @@ mod example {
                 if let Some(pid) = parts.next() {
                     let pid = nix::unistd::Pid::from_raw(pid.parse()?);
                     println!("Attaching to process {}", pid);
-                    let (remote, status) = LinuxTarget::attach(pid)?;
+                    let (remote, status) =
+                        LinuxTarget::attach(pid, AttachOptions { kill_on_exit: true })?;
```
bjorn3, line 142: Can this be falss? gdb and lldb detach on exit when they haven't spawned the process themself.

bjorn3:
Rustfmt failed

PR #66, opened 2020-07-31
Fix spelling in comments and variable names

hunger:
This makes my spellchecker happy, so that I can use it to avoid
adding more typos in the future:-)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,5 @@
 //! Headcrab, a modern Rust debugging library.
 
-#[macro_use]
-extern crate lazy_static;
-
 /// Functions to work with target processes: reading & writing memory, process control functions, etc.
 pub mod target;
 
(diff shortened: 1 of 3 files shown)


bjorn3:
First two commits LGTM. Re the third commit: I prefer mod.rs, as vscode otherwise puts it at the bottom of directory listing far away from the directory it's submodules are in.

hunger:
@bjorn3 : I never use the file system tree for navigation, I just hit Ctrl-P and start typing. It is really annoying to have lots of mod.rs files then. I also prefer symbol.rs, target.rs, etc. in my list of open documents over having mod.rs in there several times.

hunger:
I dropped the 3rd patch for now, running only with those that @bjorn3 was OK with:-) I will push the third one later separately.

bjorn3:
Thanks!

PR #68, opened 2020-08-02
Avoid lots of mod.rs files

hunger:
Use the module_name.rs name over module_name/mod.rs as that is nicer
to handle while editing.

It avoids having lots of mod.rs files in the list of open files and on tabs.


hunger:
There does not seem interest in this: Closing.

nbaksalyar:
@hunger, somehow I missed the PR, sorry!

For now, in the source root we have `src/target/linux.rs`, but `src/macos/mod.rs`, which is inconsistent. It would be definitely good to have everything in the same style, + we should [document this](https://github.com/headcrab-rs/headcrab/blob/master/CONTRIBUTING.md#coding-guidelines).

In this case, we'll also need to rename `src/windows/mod.rs` to `src/windows.rs`.
Do you want to do this as a part of this PR?

hunger:
Sure, I can do that.

PR #69, opened 2020-08-02
Support Windows Build

### LINKED ISSUE #67: Support headcrab compilation on Windows
Currently headcrab fails to compile on windows due to dependencies not being based on the platform entirely. 
The idea is  to get the code building on windows platform.

DeltaManiac:
This PR should fix #67 

Moved most of the dependencies to the unix platform


.github/workflows/windows.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Format
+      run: cargo fmt --all -- --check
+    - name: Clippy
+      run: cargo clippy
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --verbose
```
bjorn3, line 26: nit: missing trailing newline

Cargo.toml
```diff
@@ -30,6 +25,12 @@ security-framework-sys = "1.0"
 # Dependencies specific to Linux
 [target.'cfg(target_os="linux")'.dependencies]
 procfs = "0.8.0"
+[target.'cfg(any(target_os = "linux", target_os = "unix"))'.dependencies]
```
bjorn3, line 28: There is already a

```
[target.'cfg(unix)'.dependencies]
```

section above.

Cargo.toml
```diff
@@ -30,6 +25,12 @@ security-framework-sys = "1.0"
...
+nix = "0.17.0"
+libproc = "0.7.2"
+libc = "0.2.72"
+object = "0.20"
```
bjorn3, line 32: `object` should remain as dependency on all OSes I think. It is necessary for symbolication in case of remote debugging of unix targets. `gimli` is also necessary for that, so I think that one should also be moved from `cfg(unix)` to all platforms.

src/lib.rs
```diff
@@ -3,5 +3,6 @@
 /// Functions to work with target processes: reading & writing memory, process control functions, etc.
 pub mod target;
 
+#[cfg(target_os="unix,linux")]
```
bjorn3, line 6:
```suggestion
#[cfg(unix)]
```

tests/attach_readmem.rs
```diff
@@ -1,5 +1,6 @@
 //! This is a simple test to attach to already running debugee process
 
+#[cfg(target_os = "linux")]
```
bjorn3, line 3:
```suggestion
#![cfg(unix)] // Requires `fork`

//! This is a simple test to attach to already running debugee process

```

Cargo.toml
```diff
@@ -31,5 +31,6 @@ security-framework-sys = "1.0"
 [target.'cfg(target_os="linux")'.dependencies]
 procfs = "0.8.0"
 
+
```
bjorn3, line 34:
```suggestion
```

tests/attach_readmem.rs
```diff
@@ -1,5 +1,6 @@
 //! This is a simple test to attach to already running debugee process
 
+#[cfg(unix)]
```
bjorn3, line 3: Please use `#![cfg(unix)]`. This test uses `fork` which is only available on unix. While the actual test is currently `#[cfg(target_os = "linux)]`, the goal is that it will run on all unix systems.

DeltaManiac: yeah, i was bit confused because the build wasn't failing for linux or bsd

PR #73, opened 2020-08-04
Add watchpoint

### LINKED ISSUE #72: data breakpoint support
Data breakpoints ([watchpoints](https://sourceware.org/gdb/onlinedocs/gdb/Set-Watchpoints.html)) are one of the most powerful features of gdb, and have helped me to find the source of a large number of memory corruption bugs. They let you break whenever some memory is read or written to. They also allow you to break on an arbitrary expression when it becomes true, but that's a bit more of a reach goal. This is wonderful for tracking down use after frees and a wide variety of other bugs where you know that some specific memory is involved in a bug, but you don't know what code has accessed it. Breakpoints are for when you suspect some **instructions** relate to a bug. Watchpoints are for when you suspect some **data** relate to a bug.

Data breakpoints can be implemented in a pretty straightforward way by utilizing the debug registers, setting one of {DR0, DR1, DR2, DR3} to the address that you'd like to break on the access of, and [setting bitflags on DR7](https://en.wikipedia.org/wiki/X86_debug_register#DR7_-_Debug_control) to communicate what kinds of events you'd like to break on.

JJendryka:
Closes #72.
Adds watchpoints to linux target using debug registers on x86_64 architecture. Depends on #41.
- [x] Implement watchpoints
- [x] Add tests


src/target/linux.rs
```diff
@@ -56,6 +67,47 @@ pub struct AttachOptions {
     pub kill_on_exit: bool,
 }
 
+pub struct Watchpoint {
+    typ: WatchpointType,
+    addr: usize,
+    size: WatchSize,
+    on_trap: std::boxed::Box<dyn FnMut(&Self)>,
```
bjorn3, line 74: Treating this like a software breakpoint, aka returning a signal from `step`/`unpause` is more flexible, as the user of headcrab can for example allow the enduser to inspect/modify the program state before continuing.

src/target/linux.rs
```diff
@@ -13,8 +13,18 @@ use std::{
 
 lazy_static::lazy_static! {
     static ref PAGE_SIZE: usize = unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize };
+    static ref DEBUG_REG: usize = unsafe {
```
bjorn3, line 16: `DEBUG_REG_OFFSET`

src/target/linux.rs
```diff
@@ -218,6 +277,103 @@ impl LinuxTarget {
 
         Ok(tasks)
     }
+
+    #[cfg(target_arch = "x86_64")]
```
bjorn3, line 281: Please make this function available on all platforms and instead return an error on unsupported platforms.

src/target/linux.rs
bjorn3: Also this needs to be `#[cfg[target_arch="x86_64")]`.

src/target/linux.rs
```diff
@@ -218,6 +277,103 @@ impl LinuxTarget {
...
+        let size_bits: u32 = (watchpoint.size as u32) << 18 + index * 2;
+
+        let enable_bit: u32 = 1 << (2 * index);
+
+        let bit_mask: u32 = (0b11 << (2 * index)) | (0b1111 << (16 + 4 * index));
+
+        let mut dr7: u64 = 0;
+
+        unsafe {
+            self.read().read(&mut dr7, *DEBUG_REG + 7 * 4).apply()?;
```
bjorn3, line 308: This needs `PTRACE_PEEKUSER` I think, not `PTRACE_PEEK`/`read_process_vm` like `read` uses.

src/target/linux.rs
```diff
@@ -218,6 +277,103 @@ impl LinuxTarget {
...
+    #[cfg(target_arch = "x86_64")]
+    fn find_empty_watchpoint(&self) -> Option<usize> {
+        let mut empty = None;
+        for i in 0..SUPPORTED_HARDWARE_WATCHPOINTS {
+            if self.watchpoints[i].is_none() {
+                empty = Some(i);
+                break;
+            }
+        }
+        return empty;
```
bjorn3, line 375: `self.watchpoints.iter().find(|w| w.is_some())`

src/target/linux.rs
JJendryka: 🤦 I somehow missed that. Thanks!

src/target/linux.rs
JJendryka: Ok sure, will remove `on_trap`

src/target/linux.rs
```diff
@@ -696,4 +939,37 @@ mod tests {
...
+                        .expect("Couldn't attach to child");
+
+                target.set_watchpoint(Watchpoint {
+                    addr: &var as *const _ as usize,
+                    typ: WatchpointType::Write,
+                    size: WatchSize::from_usize(std::mem::size_of::<u8>())?,
+                })?;
+
+                ptrace::cont(child, Some(signal::Signal::SIGCONT)).unwrap();
+                wait::waitpid(child, None)?;
```
bjorn3, line 968: Can you check the received signal?

src/target/linux.rs
```diff
@@ -218,6 +299,168 @@ impl LinuxTarget {
...
+        &mut self,
+        watchpoint: Watchpoint,
+    ) -> Result<usize, Box<dyn std::error::Error>> {
+        #[cfg(target_arch = "x86_64")]
+        {
+            let empty = self.find_empty_watchpoint();
+            if empty.is_none() {
+                return Err(Box::new(WatchpointError::NoEmptyWatchpoint));
+            }
+            let index = empty.unwrap();
```
bjorn3, line 314:
```suggestion
            let empty = if let Some(empty) = self.find_empty_watchpoint() {
                empty
            } else {
                return Err(Box::new(WatchpointError::NoEmptyWatchpoint));
            }
```

src/target/linux.rs
```diff
@@ -218,6 +299,168 @@ impl LinuxTarget {
...
+
+            let rw_bits: u64 = match watchpoint.typ {
+                WatchpointType::Execute => 0b00,
+                WatchpointType::Read => 0b11,
+                WatchpointType::ReadWrite => 0b11,
```
bjorn3, line 319: Should these be different from each other?

src/target/linux.rs
```diff
@@ -56,6 +68,68 @@ pub struct AttachOptions {
     pub kill_on_exit: bool,
 }
 
+#[derive(Debug)]
+pub struct Watchpoint {
```
bjorn3, line 72: Maybe call this `HardwareBreakpoint`? You don't necessarily need to implement watchpoints on top of them.

src/target/linux.rs
JJendryka: Actually there is no read hardware breakpoint on x86_64. My plan is to somehow distinguish whether ReadWrite was triggered by read or write.

tests/testees/.gitignore
```diff
@@ -1,2 +1,3 @@
 /hello
 /longer_hello
+/hardware_breakpoint
```
bjorn3, line 3: nit: missing trailing newline

src/target/linux.rs
```diff
@@ -218,6 +302,193 @@ impl LinuxTarget {
...
+        {
+            if self.hardware_breakpoints[index].is_none() {
+                return Err(Box::new(HardwareBreakpointError::DoesNotExist(index)));
+            }
+
+            let mut dr7: u64;
+            let mut dr6: u64;
+
+            dr7 = self.ptrace_peekuser((*DEBUG_REG_OFFSET + 7 * 8) as *mut libc::c_void)? as u64;
+            dr6 = self.ptrace_peekuser((*DEBUG_REG_OFFSET + 6 * 8) as *mut libc::c_void)? as u64;
```
bjorn3, line 388: `let mut dr6 = ...`

src/target/linux.rs
```diff
@@ -218,6 +302,193 @@ impl LinuxTarget {
...
+
+            let bit_mask: u32 = (0b11 << (2 * index)) | (0b1111 << (16 + 4 * index));
```
bjorn3, line 390: dr7_mask, same for dr6 below.

JJendryka:
For some reason the test crashes on CI when trying to find address of static variable (not related to my PR), however I cannot reproduce the issue locally.

bjorn3:
Try using `write_volatile` to write tot the statics. Maybe LLVM was able to infer that the statics are never read again and thus the writes can be omitted, allowing the statics to be gc'ed?

bjorn3:
Rustfmt fails

JJendryka:
CI still fails, even with write_volatile. Hard to fix it without reproducing it.

bjorn3:
I pushed a commit with extra debug println's.

bjorn3:
I found the issue. Once #79 is merged, the test should pass.

bjorn3:
```
---- target::linux::tests::reads_threads stdout ----

thread 'target::linux::tests::reads_threads' panicked at 'called `Option::unwrap()` on a `None` value', src/target/linux.rs:691:23
```

One of the unwraps at the following fails.

```rust
let threads: Vec<_> = threads
    .iter()
    .map(|t| (t.name().unwrap().unwrap().clone(), t.thread_id()))
    .collect();
```

bjorn3:
@nbaksalyar Could you restart the travis job? Maybe one of the test threads exited while trying to get their name.

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
...
+                    0 as *mut libc::c_void,
+                )?;
+            }
+
+            self.hardware_breakpoints[index] = Some(breakpoint);
+
+            Ok(index)
+        }
+        #[cfg(not(target_arch = "x86_64"))]
+        Err(Box::new(HardwareBreakpointError::UnsupportedPlatform))
```
nbaksalyar, line 375: We can probably replace
```rust
#[cfg(target_arch = "x86_64")]
{
    let index = ...;
}
#[cfg(not(target_arch = "x86_64"))]
Err(Box::new(HardwareBreakpointError::UnsupportedPlatform))
```
with
```rust
if !cfg!(target_arch = "x86_64")  {
    return Err(Box::new(HardwareBreakpointError::UnsupportedPlatform));
}
let index = ...;
```
Since it will be translated into something like
```rust
if true { return Err(...) }
```
on non-x86_64 platforms, the compiler should be smart enough to eleminate unused code.

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
...
+            };
+
+            let rw_bits: u64 = match breakpoint.typ {
+                HardwareBreakpointType::Execute => 0b00,
+                HardwareBreakpointType::Read => 0b11,
+                HardwareBreakpointType::ReadWrite => 0b11,
+                HardwareBreakpointType::Write => 0b01,
+            } << 16 + index * 4;
+
+            let size_bits: u64 = (breakpoint.size as u64) << (18 + index * 4);
```
nbaksalyar, line 330: I'd suggest to move this to a function in `HardwareBreakpoint`, e.g.:

```rust
impl HardwareBreakpoint {
    fn size_bits(&self, index: usize) -> u64 {
        (self.size as u64) << (18 + index * 4)
    }
}
```
then it can be used locally as
```suggestion
            let size_bits = breakpoint.size_bits(index);
```

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
...
+
+            let enable_bit: u64 = 1 << (2 * index);
+
+            let bit_mask: u64 = (0b11 << (2 * index)) | (0b1111 << (16 + 4 * index));
```
nbaksalyar, line 334: This would also work good as a function in `HardwareBreakpoint`, e.g.:
```rust
impl HardwareBreakpoint {
    const fn bit_mask(index: usize) -> u64 {
        (0b11 << (2 * index)) | (0b1111 << (16 + 4 * index))
    }
}
```
then this line can be replaced with
```suggestion
            let bit_mask = HardwareBreakpoint::bit_mask(index);
```

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
...
+            } else {
+                return Err(Box::new(HardwareBreakpointError::NoEmptyWatchpoint));
+            };
+
+            let rw_bits: u64 = match breakpoint.typ {
+                HardwareBreakpointType::Execute => 0b00,
+                HardwareBreakpointType::Read => 0b11,
+                HardwareBreakpointType::ReadWrite => 0b11,
+                HardwareBreakpointType::Write => 0b01,
+            } << 16 + index * 4;
```
nbaksalyar, line 328: This also can be a part of `impl HardwareBreakpointType`, I think

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
...
+            let watchpoint = std::mem::replace(&mut self.hardware_breakpoints[index], None);
+            Ok(watchpoint.unwrap())
+        }
+
+        #[cfg(not(target_arch = "x86_64"))]
+        Err(Box::new(HardwareBreakpointError::UnsupportedPlatform))
+    }
+
+    pub fn clear_all_hardware_breakpoints(&mut self) -> Result<(), Box<dyn std::error::Error>> {
+        {
```
nbaksalyar, line 425: I think we don't need this to be nested? I.e. this can probably just go into the `fn` scope

src/target/linux.rs
bjorn3: `DEBUG_REG_OFFSET` only exists on x86_64.

src/target/linux.rs
```diff
@@ -221,6 +306,192 @@ impl LinuxTarget {
 
         Ok(tasks)
     }
+
+    #[cfg(target_arch = "x86_64")]
```
bjorn3, line 310:
```suggestion
```

nbaksalyar:
Rustfmt seems to be failing

src/target/linux.rs
```diff
@@ -221,6 +245,181 @@ impl LinuxTarget {
...
+            if dr7 & (1 << (2 * index)) != 0 {
+                // Panic for now
+                panic!("Invalid debug register state")
+            }
+
+            dr7 = (dr7 & !bit_mask) | (enable_bit | rw_bits | size_bits);
+
+            #[allow(deprecated)]
+            unsafe {
+                //Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```
bjorn3, line 279:
```suggestion
                // Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```

nit

src/target/linux.rs
```diff
@@ -221,6 +245,181 @@ impl LinuxTarget {
...
+
+            let dr7_bit_mask: u64 = HardwareBreakpoint::bit_mask(index);
+            dr7 = dr7 & !dr7_bit_mask;
+
+            let dr6_bit_mask: u64 = 1 << index;
+            dr6 = dr6 & !dr6_bit_mask as u64;
+
+            #[allow(deprecated)]
+            unsafe {
+                //Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```
bjorn3, line 331:
```suggestion
                // Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```

nit

src/target/linux.rs
```diff
@@ -221,6 +245,181 @@ impl LinuxTarget {
...
+    ) -> Result<Option<usize>, Box<dyn std::error::Error>> {
+        #[cfg(target_arch = "x86_64")]
+        {
+            let mut dr7 = self.ptrace_peekuser((*DEBUG_REG_OFFSET + 6 * 8) as *mut libc::c_void)?;
+
+            for i in 0..SUPPORTED_HARDWARE_BREAKPOINTS {
+                if dr7 & (1 << i) != 0 && self.hardware_breakpoints[i].is_some() {
+                    // Clear bit for this breakpoint
+                    dr7 &= !(1 << i);
+                    //Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```
bjorn3, line 377:
```suggestion
                    // Have to use deprecated function because of no alternative for PTRACE_POKEUSER
```

nit

bjorn3:
Thanks!

PR #74, opened 2020-08-05
Symbol demangling

### LINKED ISSUE #71: Demangle symbol names
We can use [rustc-demangle](https://github.com/alexcrichton/rustc-demangle/) for demangling of Rust symbols.
Since it's more of a presentation problem, we should also save the original name, so it might look like this:

```rust
use rustc_demangle::demangle;

struct Symbol<'a> {
  orig_name: &'a str,
}

impl<'a> Symbol<'a> {
  fn get_name(&self) -> String { demangle(self.orig_name).to_string() }
  fn get_orig_name(&self) -> &str { self.orig_name }
}
```

The Rust-specific implementation should probably be confined to a Rust lang module (in case if we want to support more languages in the future).

Stupremee:
I'm not sure if this is the approach you wanted, but I think this should work.

Resolves #71

### DIFF
--- a/src/symbol/mod.rs
+++ b/src/symbol/mod.rs
@@ -283,7 +283,11 @@ impl Dwarf {
     }
 
     pub fn get_address_symbol_name(&self, addr: usize) -> Option<String> {
-        self.rent(|parsed| Some(parsed.get_address_symbol(addr)?.name()?.to_string()))
+        self.rent(|parsed| Some(parsed.get_address_symbol(addr)?.orig_name()?.to_string()))
+    }
+
+    pub fn get_address_demangled_name(&self, addr: usize) -> Option<String> {
+        self.rent(|parsed| Some(parsed.get_address_symbol(addr)?.name()?))
     }
 
     pub fn get_address_symbol_kind(&self, addr: usize) -> Option<SymbolKind> {
@@ -409,6 +413,18 @@ impl RelocatedDwarf {
             return entry
                 .dwarf
                 .get_address_symbol_name(addr - entry.bias as usize);
+        }
+        None
+    }
+
+    pub fn get_address_demangled_name(&self, addr: usize) -> Option<String> {
+        for entry in &self.0 {
+            if (addr as u64) < entry.address_range.0 || addr as u64 >= entry.address_range.1 {
+                continue;
+            }
+            return entry
+                .dwarf
+                .get_address_demangled_name(addr - entry.bias as usize);
         }
         None
     }
--- a/src/symbol/mod.rs
+++ b/src/symbol/mod.rs
@@ -1,11 +1,12 @@
 //! This module provides a naive implementation of symbolication for the time being.
 //! It should be expanded to support multiple data sources.
 
+mod sym;
 pub mod unwind;
 
 use gimli::read::{EvaluationResult, Reader as _};
 use object::{
-    read::{Object, ObjectSection, ObjectSegment, Symbol},
+    read::{Object, ObjectSection, ObjectSegment},
     SymbolKind,
 };
 use std::{
@@ -15,6 +16,7 @@ use std::{
     path::Path,
     rc::Rc,
 };
+pub use sym::Symbol;
 
 mod source;
 
@@ -157,8 +159,9 @@ impl<'a> ParsedDwarf<'a> {
                 }
                 !symbol.is_undefined() && symbol.section() != object::SymbolSection::Common
             })
+            .map(Into::into)
             .collect();
-        symbols.sort_by_key(|sym| sym.address());
+        symbols.sort_by_key(|sym: &Symbol| sym.address());
 
         let mut symbol_names = HashMap::new();
         for sym in &symbols {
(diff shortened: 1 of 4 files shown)


Stupremee:
There's definitely something not working. Please dont merge

nbaksalyar:
Thanks, this is great! I think we can also change the list of expected names in [tests/unwind_stack.rs](https://github.com/headcrab-rs/headcrab/blob/2b8dc1bf744b395b5e730228aec3be9a708f6344/tests/unwind_stack.rs#L51-L56) to verify the demangling routines work correctly. Do you want to do this as a part of this pull request?

Stupremee:
Yes that's how I saw that my code is broken

bjorn3:
I just remembered that `addr2line` already has demangling support. This also handles inlined functions. https://docs.rs/addr2line/0.13.0/addr2line/struct.FunctionName.html#method.demangle

Directly using `rustc-demangle` is useful for when there is no DWARF debuginfo though.

tests/unwind_stack.rs
```diff
@@ -50,9 +50,9 @@ fn unwind_stack() -> Result<(), Box<dyn std::error::Error>> {
 
     let expected = &[
         "breakpoint",
-        "_ZN5hello4main17h",
-        "_ZN3std2rt10lang_start28_$u7b$$u7b$closure$u7d$$u7d$17h",
-        "_ZN3std2rt19lang_start_internal17h",
+        "hello::main::",
+        "std::rt::lang_start::{{closure}}::",
+        "std::rt::lang_start_internal::",
```
bjorn3, line 55: I think `get_address_symbol_name` should not demangle the name. Instead there could be an extra method `get_address_demangled_name`.

nbaksalyar:
> `addr2line` already has demangling support

Yeah, it also seems to be automatically choosing between the C++ & Rust mangling schemes depending on DWARF lang param, which is a nice feature if we want to support more langs in the future. So we can probably use addr2line as the default demangling function and resort to the direct `rustc-demangle` when there's no DWARF available?

Stupremee:
> I just remembered that addr2line already has demangling support. 

That sounds good. Should I wait for #54 or add it here as well?

bjorn3:
Please wait, it conflicts with #54.

nbaksalyar:
#54 is merged now :+1:

src/symbol/sym.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Returns the unmangled name of this symbol.
+    #[inline]
+    pub fn orig_name(&self) -> Option<&'data str> {
+        self.symbol.name()
+    }
+
+    /// Return the kind of this symbol.
+    #[inline]
+    pub fn kind(&self) -> SymbolKind {
+        self.symbol.kind()
```
nbaksalyar, line 32: I'd suggest adding `impl Deref<Target = object::Symbol>` to replace functions which have 1-to-1 mappings.
I.e., something like this:
```rust
impl Deref for Symbol {
    type Target = object::Symbol;

    fn deref(&self) -> &Self::Target {
        &self.symbol
    }
}
```

src/symbol/sym.rs
```diff
@@ -0,0 +1,48 @@
...
+/// A symbol table entry.
+#[derive(Clone, Debug)]
+pub struct Symbol<'data> {
+    symbol: object::Symbol<'data>,
+}
+
+impl<'data> Symbol<'data> {
+    /// Returns the demangled name if this symbol has a name.
+    #[inline]
+    pub fn name(&self) -> Option<String> {
```
bjorn3, line 17:
```suggestion
    pub fn demangled_name(&self) -> Option<String> {
```

Then `orig_name` is not necessary anymore.

src/symbol/mod.rs
```diff
@@ -157,13 +159,14 @@ impl<'a> ParsedDwarf<'a> {
...
             .collect();
-        symbols.sort_by_key(|sym| sym.address());
+        symbols.sort_by_key(|sym: &Symbol| sym.address());
 
         let mut symbol_names = HashMap::new();
         for sym in &symbols {
             if let Some(name) = sym.name() {
-                symbol_names.insert(name, sym.address() as usize);
+                symbol_names.insert(name.to_string(), sym.address() as usize);
             }
```
bjorn3, line 170:
```suggestion
            }
            if let Some(name) = sym.demangled_name() {
                symbol_names.insert(name, sym.address() as usize);
            }
```

src/symbol/sym.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Clone, Debug)]
+pub struct Symbol<'data> {
+    symbol: object::Symbol<'data>,
+}
+
+impl<'data> Symbol<'data> {
+    /// Returns the demangled name if this symbol has a name.
+    #[inline]
+    pub fn demangled_name(&self) -> Option<String> {
+        // TODO: Avoid this allocation in every call. (lifetime errors)
```
bjorn3, line 18: You can store the demangled name in field and then return a reference.

bjorn3:
Can you change https://github.com/headcrab-rs/headcrab/blob/25b85b7fb8129e3368ec62e0f5fc9ab89f80091a/examples/repl.rs#L230 to use the demangled name?

nbaksalyar:
Thank you!

PR #75, opened 2020-08-07
ReadMem improvements

nbaksalyar:
This PR cherry-picks some improvements from #41 (to leave it write-specific and more manageable to review).

It also introduces new methods, `ReadMem::read_slice` and `ReadMem::read_ptr`, which can be used for reading vectors and arrays from debuggee's memory.


src/target/linux/memory.rs
```diff
@@ -0,0 +1,36 @@
...
+        maps.binary_search_by(|map| {
+            if op.remote_base() < map.address.0 as usize {
+                Ordering::Greater
+            } else if op.remote_base() > map.address.1 as usize {
+                Ordering::Less
+            } else {
+                Ordering::Equal
+            }
+        })
+        .is_ok()
```
bjorn3, line 32: This should check `is_readable`.

src/target/linux/memory.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// Abstract memory operation (reading or writing).
+pub trait MemoryOp {
+    /// Returns a remote address at which this memory operation will be applied to.
+    fn remote_base(&self) -> usize;
+}
+
+/// Splits memory operations to those that read from read protected memory and those that do not.
+pub fn split_protected<'a, M: MemoryOp + Clone>(
```
bjorn3, line 18: This is also necessary for write protected memory.

src/target/linux/readmem.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// The provided value `val` can't be accessed until either `apply` is called or `self` is
+    /// dropped.
+    ///
+    /// # Safety
+    ///
+    /// The type `T` must not have any invalid values.
+    /// For example, `T` must not be a `bool`, as `transmute::<u8, bool>(2)` is not a valid value for a bool.
+    /// In case of doubt, wrap the type in [`mem::MaybeUninit`].
+    // todo: further document mem safety - e.g., what happens in the case of partial read
+    pub unsafe fn read_slice<T>(mut self, val: &'a mut [T], remote_base: usize) -> Self {
```
bjorn3, line 86: Can we also have a safe `read_byte_slice` method?

src/target/linux/readmem.rs
```diff
@@ -0,0 +1,277 @@
...
+            let (protected, readable) = split_protected(&protected_maps, &self.read_ops)?;
+
+            self.read_process_vm(&readable)?;
+            self.read_ptrace(&protected)?;
+        }
+        Ok(())
+    }
+
+    /// Allows to read from several different locations with one system call.
+    /// It will ignore pages that are not readable. Returns number of bytes read at granularity of ReadOps.
```
bjorn3, line 132:
```suggestion
    /// It will error on pages that are not readable. Returns number of bytes read at granularity of ReadOps.
```

src/target/linux/readmem.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Converts the memory read operation into a local IoVec.
+    fn as_local_iovec(&self) -> libc::iovec {
+        libc::iovec {
+            iov_base: self.local_ptr,
+            iov_len: self.len,
+        }
+    }
+
+    /// Splits ReadOp so that each resulting ReadOp resides in only one memory page.
+    fn split_on_page_boundary(&self) -> Vec<ReadOp> {
```
bjorn3, line 240: Maybe take `&mut Vec<ReadOp>` to push directly into?

src/target/linux/memory.rs
nbaksalyar: Good catch, thanks - I made it generic but haven't changed the comments and variable names for some reason :sweat_smile:

bjorn3: It should check for `is_writable` in the write case.

src/target/linux/memory.rs
nbaksalyar: It does, but in an implicit way - `maps` are pre-filtered as `memory_maps().into_iter().filter(|m| !m.is_readable())` at the call site. I'm not sure if we should make it a func argument for explicit filtering, or leave it as a comment suggesting that it's a user's responsibility.

bjorn3: Please take an `Iterator` instead of slice then to avoid an unnecessary allocation and explain in the doc comment that the caller is responsible for filtering.

src/target/linux/memory.rs
```diff
@@ -0,0 +1,44 @@
...
+/// This function can be used for both write or read operations, and `maps` should be pre-filtered
+/// to contain only protected pages, e.g.:
+/// ```
+/// use headcrab::target::MemoryMap;
+///
+/// let maps: Vec<MemoryMap> = vec![];
+/// let protected_maps = maps.into_iter().filter(|map| !map.is_writable);
+/// ```
+pub fn split_protected<'a, M: MemoryOp>(
+    maps: &'a [MemoryMap],
```
bjorn3, line 27: Maybe call this `protected_maps`?

src/target/linux/memory.rs
bjorn3: I actually meant for `maps` to be an iterator.

src/target/linux/readmem.rs
```diff
@@ -0,0 +1,287 @@
...
+        {
+            let protected_maps = self
+                .target
+                .memory_maps()?
+                .into_iter()
+                .filter(|map| !map.is_readable)
+                .collect::<Vec<_>>();
+
+            let (protected, readable) =
+                split_protected(&protected_maps, self.read_ops.iter().cloned())?;
```
bjorn3, line 136: You can use `self.read_ops.into_iter()` because `self` is passed by-value.

nbaksalyar: > for maps to be an iterator.

Hmm, I guess then we'd need to convert it into a slice anyway, as `binary_search_by` is a `std::slice` method.
Or did you mean something else?

bjorn3: :facepalm:

nbaksalyar:
Thanks for the review!

PR #76, opened 2020-08-07
Windows Target

### LINKED ISSUE #77: Add Windows Target
A new windows target to launch a debuggee and attach to a running process

DeltaManiac:
This PR adds the windows target to headcrab. 
Fixes #77 
- [x] Launch a debuggee process
- [x] Attach to a running process

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ libc = "0.2.72"
 # Dependencies specific to macOS
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 mach = "0.3"
-
 security-framework-sys = "1.0"
 
 # Dependencies specific to Linux
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,11 +28,15 @@ libc = "0.2.72"
 # Dependencies specific to macOS
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 mach = "0.3"
+
 security-framework-sys = "1.0"
 
 # Dependencies specific to Linux
 [target.'cfg(target_os="linux")'.dependencies]
 procfs = "0.8.0"
+
+[target.'cfg(target_os="windows")'.dependencies]
+winapi = { version = "0.3.9", features = ["winuser","processthreadsapi","winbase","minwinbase","debugapi","winnt","memoryapi","dbghelp"] }
 
 [dev-dependencies]
 rustyline = "6.2.0"
--- a/src/target/windows/mod.rs
+++ b/src/target/windows/mod.rs
@@ -39,7 +39,7 @@ impl Target {
                 std::ptr::null_mut(),
                 std::ptr::null_mut(),
                 &mut startup_info,
-                &mut proc_info
+                &mut proc_info,
             )
         } == FALSE
         {
--- a/src/target/windows/mod.rs
+++ b/src/target/windows/mod.rs
@@ -42,15 +42,6 @@ impl Target {
         {
             return Err(Box::new(std::io::Error::last_os_error()));
         }
-        // Initialize the SymbolHandler
-        unsafe {
-            dbghelp::SymSetOptions(
-                dbghelp::SymGetOptions() | dbghelp::SYMOPT_DEBUG | dbghelp::SYMOPT_LOAD_LINES,
-            )
-        };
-        if unsafe { dbghelp::SymInitializeW(pi.hProcess, ptr::null(), FALSE) } == FALSE {
-            return Err(Box::new(std::io::Error::last_os_error()));
-        }
 
         Ok(Target { proc_info: pi })
     }
--- a/src/target/windows/mod.rs
+++ b/src/target/windows/mod.rs
@@ -1,13 +1,14 @@
 use std::mem;
-use std::ptr;
 use winapi::shared::minwindef::FALSE;
-use winapi::um::dbghelp;
-use winapi::um::processthreadsapi::{CreateProcessW, PROCESS_INFORMATION, STARTUPINFOW};
+use winapi::um::processthreadsapi::{
+    CreateProcessW, OpenProcess, PROCESS_INFORMATION, STARTUPINFOW,
+};
 use winapi::um::winbase;
+use winapi::um::winnt;
 
 /// This structure holds the state of the debuggee on windows systems
 pub struct Target {
-    proc_info: PROCESS_INFORMATION,
+    process_handle: winnt::HANDLE,
 }
 
 macro_rules! wide_string {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -28,11 +28,15 @@ libc = "0.2.72"
 # Dependencies specific to macOS
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 mach = "0.3"
+
```
bjorn3, line 31:
```suggestion
```

nit

src/target/windows/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            return Err(Box::new(std::io::Error::last_os_error()));
+        }
+        // Initialize the SymbolHandler
+        unsafe {
+            dbghelp::SymSetOptions(
+                dbghelp::SymGetOptions() | dbghelp::SYMOPT_DEBUG | dbghelp::SYMOPT_LOAD_LINES,
+            )
+        };
+        if unsafe { dbghelp::SymInitializeW(pi.hProcess, ptr::null(), FALSE) } == FALSE {
```
bjorn3, line 51: https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitialize

> A process that calls SymInitialize should not call it again unless it calls SymCleanup first. If the call to SymInitialize set fInvadeProcess to TRUE and you simply need to reload the module list, use the SymRefreshModuleList function.

> All DbgHelp functions, such as this one, are single threaded. Therefore, calls from more than one thread to this function will likely result in unexpected behavior or memory corruption. To avoid this, call SymInitialize only when your process starts and SymCleanup only when your process ends. It is not necessary for each thread in the process to call these functions.

This isn't safe to do. You would need to make `Target` neither `Send` nor `Sync`. You also need a static to ensure that `Target` is dropped before a new one is created and call `SymCleanup` from the drop implementation of `Target`.

I think using https://github.com/willglynn/pdb as suggested by @nbaksalyar is better.

DeltaManiac: Thanks @bjorn3  
Yeah I did initially think i would have to use the `pdb` crate but then cancelled out that plan once i read the msdn docs and thought "Winapi has an inbuilt symbol loader. Why not use that."

src/target/windows/mod.rs
```diff
@@ -0,0 +1,61 @@
+use std::mem;
+use winapi::shared::minwindef::FALSE;
+use winapi::um::processthreadsapi::{
+    CreateProcessW, OpenProcess, PROCESS_INFORMATION, STARTUPINFOW,
+};
+use winapi::um::winbase;
+use winapi::um::winnt;
+
+/// This structure holds the state of the debuggee on windows systems
```
bjorn3, line 9: nit:
```suggestion
/// This structure holds the state of the debuggee on windows systems.
```

src/target/windows/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    ($string:expr) => {{
+        use std::os::windows::ffi::OsStrExt;
+        let input = std::ffi::OsStr::new($string);
+        let vec: Vec<u16> = input.encode_wide().chain(Some(0)).collect();
+        vec
+    }};
+}
+
+impl Target {
+    /// Launch a new debuggee process
```
bjorn3, line 24: nit:
```suggestion
    /// Launch a new debuggee process.
```

src/target/windows/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            return Err(Box::new(std::io::Error::last_os_error()));
+        }
+
+        Ok(Target {
+            process_handle: pi.hProcess,
+        })
+    }
+
+    /// Attach to a running Process
```
bjorn3, line 52: nit:
```suggestion
    /// Attach to a running Process.
```

DeltaManiac:
@bjorn3 
I wonder why `cargo fmt` doesnt fix the nit. 
Is there any config ?

bjorn3:
Rustfmt doesn't know anything about doc comment contents.

src/target/windows/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        let input = std::ffi::OsStr::new($string);
+        let vec: Vec<u16> = input.encode_wide().chain(Some(0)).collect();
+        vec
+    }};
+}
+
+impl Target {
+    /// Launch a new debuggee process.
+    pub fn launch(path: &str) -> Result<Target, Box<dyn std::error::Error>> {
+        let mut si: STARTUPINFOW = unsafe { mem::zeroed() };
```
nbaksalyar, line 26: I'd suggest using [MaybeUninit](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html) here to explicitly tell the compiler that this value is not initialised and is not to be used.

So it can be used as e.g.
```rust
let mut si: STARTUPINFOW = MaybeUninit::uninit();
let mut pi: PROCESS_INFORMATION = MaybeUninit::uninit();

CreateProcessW(
    ...
    si.as_mut_ptr(),
    pi.as_mut_ptr(),
);
```

src/target/windows/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        let mut pi: PROCESS_INFORMATION = unsafe { mem::zeroed() };
```
nbaksalyar, line 27: Nit: even though it's clear what these variables represent in the context, I think it's still better to use expanded names, e.g.
```suggestion
        let mut startup_info: STARTUPINFOW = unsafe { mem::zeroed() };
        let mut proc_info: PROCESS_INFORMATION = unsafe { mem::zeroed() };
```

src/target/windows/mod.rs
bjorn3: `STARTUPINFOW` is an input value, not an output value.

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

> The caller is responsible for ensuring that the standard handle fields in STARTUPINFO contain valid handle values.

nbaksalyar: Makes sense then, thanks!

DeltaManiac: Shouldn't i use the `zeroed` function of `MaybeUninit`?
This
```rust
let mut si: STARTUPINFOW = mem::MaybeUninit::uninit();
```
be more like
```rust
let mut si= mem::MaybeUninit::<STARTUPINFOW>::zeroed();
let mut si = unsafe{si.assume_init()};
```
`mem::MaybeUninit::uninit()` requires us to `drop` explicitly.

bjorn3: The first member of `STARTUPINFO` is the size of the whole struct.

src/target/windows/mod.rs
DeltaManiac: Done

DeltaManiac:
@nbaksalyar Can this be merged now ?

PR #81, opened 2020-08-10
Make threads test more robust

### LINKED ISSUE #80: Make threads test robust for disappearing threads
To prevent https://github.com/headcrab-rs/headcrab/pull/73#issuecomment-670928771

fabiim:
Fixes #80  

Can handle the failure to obtain the name of a thread. This might
happen for different reasons (e.g., maybe the thread is not there
anymore). I'm keeping the error as the name to show in the assertions
failures.

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -417,7 +417,14 @@ mod tests {
 
         let threads: Vec<_> = threads
             .iter()
-            .map(|t| (t.name().unwrap().unwrap().clone(), t.thread_id()))
+            .map(|t| {
+                let name = match t.name() {
+                    Ok(Some(name)) => name,
+                    Ok(None) => "".to_string(),
+                    Err(err) => err.to_string(),
+                };
+                (name, t.thread_id())
+            })
             .collect();
 
         assert!(
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -418,7 +418,7 @@ mod tests {
         let threads: Vec<_> = threads
             .iter()
             .map(|t| {
-                let name = t.name().unwrap().unwrap_or("".to_string());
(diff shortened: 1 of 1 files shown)


src/target/linux.rs
```diff
@@ -417,7 +417,14 @@ mod tests {
 
         let threads: Vec<_> = threads
             .iter()
-            .map(|t| (t.name().unwrap().unwrap().clone(), t.thread_id()))
+            .map(|t| {
+                let name = match t.name() {
+                    Ok(Some(name)) => name,
+                    Ok(None) => "".to_string(),
+                    Err(err) => err.to_string(),
```
bjorn3, line 424: I think only `Ok(Some(_))` and `Ok(None)` can happen by chance without there actually being a problem in either headcrab or the kernel. I think `Err(_)` would indicate a bug.

fabiim: I see. Do you think I should just unwrap the result without worries?

bjorn3: Yes

fabiim: Thank you! Fixed.

fabiim:
I can't figure out from [here](https://cirrus-ci.com/task/5452032419037184) why is it failing...

src/target/linux.rs
```diff
@@ -417,7 +417,11 @@ mod tests {
 
         let threads: Vec<_> = threads
             .iter()
-            .map(|t| (t.name().unwrap().unwrap().clone(), t.thread_id()))
+            .map(|t| {
+                let name = t.name().unwrap().unwrap_or("".to_string());
```
bjorn3, line 421:
```suggestion
                let name = t.name().unwrap().unwrap_or_else(String::new);
```

Prevents an allocation in the happy case.

PR #85, opened 2020-08-16
Get a thread list from an OSX task

fabiim:
- Using the thread port as the id because we will need it to
control the thread (as in [delve](https://github.com/go-delve/delve/blob/af250d45126070adbc702fe4ae1b74f19dcfdf97/pkg/proc/native/proc_darwin.go#L260)).
- Most mach threads belong to a pthread from which we can get the name.
Couldn't find any other way. This doesn't resolve the main thread name.


src/target/macos.rs
```diff
@@ -19,6 +26,53 @@ use std::{
...
+            let mut name = [0 as libc::c_char; MAX_THREAD_NAME];
+            let name_ptr = &mut name as *mut [libc::c_char] as *mut libc::c_char;
+            let get_name = unsafe { libc::pthread_getname_np(pt_id, name_ptr, MAX_THREAD_NAME) };
+            if get_name == 0 {
+                let opt_name = unsafe { CStr::from_ptr(name_ptr) }
+                    .to_str()
+                    .map(|s| s.to_owned())
+                    .ok();
+                return Ok(opt_name);
+            }
```
bjorn3, line 66: This should return `Err(...)` in the else case.

src/target/macos.rs
```diff
@@ -19,6 +26,53 @@ use std::{
...
+    fn name(&self) -> Result<Option<String>, Box<dyn Error>> {
+        if let Some(pt_id) = self.pthread_id {
+            let mut name = [0 as libc::c_char; MAX_THREAD_NAME];
+            let name_ptr = &mut name as *mut [libc::c_char] as *mut libc::c_char;
+            let get_name = unsafe { libc::pthread_getname_np(pt_id, name_ptr, MAX_THREAD_NAME) };
+            if get_name == 0 {
+                let opt_name = unsafe { CStr::from_ptr(name_ptr) }
+                    .to_str()
+                    .map(|s| s.to_owned())
+                    .ok();
```
bjorn3, line 64: This should return an error when the string is not valid utf-8 rather than `Ok(None)`.

src/target/macos.rs
```diff
@@ -19,6 +26,53 @@ use std::{
...
+                return Ok(opt_name);
+            }
+        };
+        Ok(None)
```
bjorn3, line 68: When there is no pthread existing for the task this should indeed return `Ok(None)`.

src/target/macos.rs
```diff
@@ -114,6 +168,45 @@ impl Target {
...
+                    port,
+                    pthread_id,
+                    task_port,
+                }) as Box<dyn Thread<ThreadId = mach_port_t>>;
+
+                osx_threads.push(thread);
+            }
+            Ok(osx_threads)
+        } else {
+            Err(format!("Failure to read task {} threads.", self.port).into())
```
bjorn3, line 207:
```suggestion
            Err(format!("Failure to read task {} threads. Error: {}", self.port, result).into())
```

src/target/macos.rs
```diff
@@ -114,6 +168,45 @@ impl Target {
...
+        let result = unsafe { mach::task::task_threads(self.port, &mut threads, &mut tcount) };
+
+        if result == kern_return::KERN_SUCCESS {
+            let tcount = tcount as usize;
+            let mut osx_threads = Vec::with_capacity(tcount);
+
+            for i in 0..tcount {
+                let port = unsafe { *threads.add(i) };
+                let pthread_id = Some(unsafe { pthread_from_mach_thread_np(port) })
+                    .filter(|pthread| *pthread != 0);
```
bjorn3, line 195:
```suggestion
                let pthread_id = match unsafe { pthread_from_mach_thread_np(port) } {
                    0 => None,
                    id => Some(id)
                };
```

src/target/macos.rs
fabiim: fixed

src/target/macos.rs
fabiim: fixed

src/target/macos.rs
```diff
@@ -19,6 +26,56 @@ use std::{
...
+                Ok(Some(name))
+            } else {
+                Err(format!(
+                    "Failure to read pthread {} name. Error: {}",
+                    pt_id, get_name
+                )
+                .into())
+            };
+        };
+        Ok(None)
```
bjorn3, line 71:
```suggestion
            if get_name == 0 {
                let name = unsafe { CStr::from_ptr(name_ptr) }.to_str()?.to_owned();
                Ok(Some(name))
            } else {
                Err(format!(
                    "Failure to read pthread {} name. Error: {}",
                    pt_id, get_name
                )
                .into())
            }
        } else {
            Ok(None)
        }
```

src/target/macos.rs
```diff
@@ -19,6 +26,56 @@ use std::{
...
+    fn drop(&mut self) {
+        let result = unsafe { mach::mach_port::mach_port_deallocate(self.task_port, self.port) };
+        if result != kern_return::KERN_SUCCESS {
+            panic!("Failed to deallocate port!");
+        }
+    }
+}
+
+extern "C" {
+    // FIXME: Use libc  > 0.2.74 when available
```
bjorn3, line 48: You mean this will be in the next release of the `libc` crate?

fabiim: Yes.

PR #93, opened 2020-08-23
Add package encoding and sending

JJendryka:
Implements GDB Remote Serial Protocol Client. It will allow to use all already existing servers as debugging targets.
- [x]  Encoding packets
- [ ] Decoding packets
- [ ] Reading registers
- [ ] Writing registers
- [ ] Reading memory
- [ ] Writing memory
- [ ] Breakpoints
-...

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -18,6 +18,9 @@ mod windows;
 #[cfg(target_os = "windows")]
 pub use windows::*;
 
+mod gdb_remote;
+pub use gdb_remote::*;
+
 mod thread;
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


luser:
FYI, I have a WIP GDB remote protocol implementation in Rust here: https://github.com/luser/rust-gdb-remote-protocol/
It's mostly focused on implementing the server side of the protocol, but it might be an interesting thing to try integrating because it'd let you punt on all of the debug information and unwinding bits for a while by letting gdb deal with all of that.

khuey:
Similarly you might find https://github.com/khuey/gdb-agent useful.

PR #95, opened 2020-08-23
Use more robust IPC mechanism for WriteMem tests

### LINKED ISSUE #92: Tests hang after target::linux::tests::read_memory ... ok
When trying to run tests they hang after executing target::linux::tests::read_memory test. It doesn't state which test will be run next, and read_memory test(last run) has already exited. I cloned a repo once again and the problem persists.

nbaksalyar:
There was a race between `SIGSTOP` and `SIGCONT` signals used for synchronization between a child & a parent process in writemem tests. This PR replaces the synchronization mechanism with pipes. Closes #92.

### DIFF
--- a/src/target/linux/writemem.rs
+++ b/src/target/linux/writemem.rs
@@ -233,7 +233,7 @@ mod tests {
             mman::{mprotect, ProtFlags},
             ptrace, wait,
         },
-        unistd::{self, fork, getppid, ForkResult},
+        unistd::{fork, ForkResult},
     };
     use std::{
         alloc::{alloc_zeroed, dealloc, Layout},
@@ -274,16 +274,22 @@ mod tests {
         let write_var2_op: u8 = 0;
         let write_array = [0u8; 4];
 
-        // Use a pipe for synchronization between parent & child processes.
-        let (read_pipe, write_pipe) = unistd::pipe().unwrap();
-
-        // ptrace::attach() is not allowed to be called on its own process, so we do a fork.
-        // child process writes to the parent's memory instead of the other way around because it's easier to
-        // check results with assert_eq! this way.
         match fork() {
             Ok(ForkResult::Child) => {
-                let parent = getppid();
-                let (target, _wait_stat) = LinuxTarget::attach(parent, Default::default()).unwrap();
+                ptrace::traceme().unwrap();
+
+                // Catch the panic so that we can report back to the original process.
+                let test_res = std::panic::catch_unwind(|| unsafe {
+                    assert_eq!(ptr::read_volatile(&write_var_op), var);
+                    assert_eq!(ptr::read_volatile(&write_var2_op), var2);
+                    assert_eq!(&ptr::read_volatile(&write_array), dyn_array.as_slice());
+                });
+
+                // Return an explicit status code.
+                std::process::exit(if test_res.is_ok() { 0 } else { 100 });
+            }
+            Ok(ForkResult::Parent { child, .. }) => {
+                let (target, _wait_stat) = LinuxTarget::attach(child, Default::default()).unwrap();
 
                 // Write memory to parent's process
                 let write_mem = target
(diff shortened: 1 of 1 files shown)


nbaksalyar:
Travis CI failed (although not visible on GitHub):
https://travis-ci.org/github/headcrab-rs/headcrab/jobs/720437235#L444-L453

Fixing this now.

src/target/linux/writemem.rs
```diff
@@ -291,12 +294,17 @@ mod tests {
...
+
                 ptrace::detach(parent, None).unwrap();
             }
             Ok(ForkResult::Parent { .. }) => {
                 // Sleep & wait for the child to send us SIGCONT
-                signal::raise(signal::Signal::SIGSTOP).unwrap();
+                unistd::close(write_pipe).unwrap();
+
+                let mut dummy = [0u8; 1];
+                let _ = unistd::read(read_pipe, &mut dummy).unwrap();
```
bjorn3, line 307: You can use `&mut [0]` directly as argument.

bjorn3 (approved):

bjorn3:
Missed https://github.com/headcrab-rs/headcrab/pull/95#issuecomment-678809658

bjorn3:
Same problem.

nbaksalyar:
> Same problem.

Yeah, Travis CI fails with a weird problem which looks like a missing SYS_PTRACE capability -- I just tried it with Docker locally, and it has the same failure, unless I add a `--cap-add=SYS_PTRACE` argument. Which is weird though, because the `readmem::read_memory` test also fails without the `SYS_PTRACE` capability. Looking into it.

.travis.yml
```diff
@@ -3,6 +3,7 @@ os:
   - osx
 
 language: rust
+sudo: true
```
bjorn3, line 6: Travis has switched from containers to vm's since spectre/meltdown, so `sudo` doesn't do anything anymore.

nbaksalyar: Yeah, I reverted the commit now.

It's also hard to understand the cause of the failure -- I contacted the Travis support, asking them to enable the debug mode for our repo.

For now, it seems the only option is to `#[ignore]` these tests temporarily. If there's no way to make it work with Travis, we could look into self-hosted GitHub Action runners. What do you think?

bjorn3: I didn't know that travis had a debug mode.

Does Github Actions also prevent this ptrace usage?

nbaksalyar: Yes, ptrace doesn’t work at all with GH Actions by default and seemingly there’s no way to elevate privileges (at least, I didn’t find any — this repo was setup with Linux GH Actions at first, and even process_vm_read didn’t work properly). There’s an option to use our own custom GH Actions runner which also allows to choose an operating system and install any software needed, so we might want to explore this in the future.

nbaksalyar:
Debug mode is now enabled for our repo, so I'll try to troubleshoot the Travis failure now.

nbaksalyar:
Found it, it's a legit failure: [`ptrace_scope`](https://www.kernel.org/doc/Documentation/security/Yama.txt) is set to 1 on Travis VMs, so ptrace is in a restricted mode. Long story short, `EPERM` here is caused by a child process trying to attach to the parent without calling `ptrace::traceme()` in the parent process first. We can also employ a more robuts error reporting in this case by writing an error to the pipe. Fix is on the way. :)

nbaksalyar:
So, ultimately this could've been solved in a simpler way: attaching a parent to a child process, as normal. The only problem I had with this is that failed `assert!`s weren't visible in the parent process, and the test was displayed as if it passed. And for some reason, panicking `assert!` doesn't set a different exit code, so we couldn't discern if it is actually failed. In order to fix that, I use `panic::catch_unwind` and now the error code is returned explicitly. Which feels kind of weird and kludge-y, but it works for now and it should unblock CI.

I also tried to use read-only mmap to share memory between the parent & the child process, but here's an interesting tidbit: [ptrace doesn't work](https://yarchive.net/comp/linux/ptrace_mmap.html) with `mmap(PROT_READ)` without `PROT_WRITE`! I.e., it behaves as if it can't write to protected memory. I guess it's also something to keep in mind for the future.

bjorn3 (approved):

Issue #87, opened 2020-08-17
Add a trait for registers


bjorn3:
Something like

```rust
trait Registers {
    fn ip(&self) -> u64;
    fn set_ip(&mut self, ip: u64);
    fn sp(&self) -> u64;
    fn set_sp(&mut self, sp: u64);
    fn bp(&self) -> Option<u64>; // None if the standard abi on the platform has no base pointer
    #[must_use]
    fn set_bp(&mut self, bp: u64) -> Option<()>; // None if the standard abi on the platform has no base pointer

    fn reg_for_dwarf(&self, reg: gimli::Register) -> Option<u64>; // None when register doesn't exist
    #[must_use]
    fn set_reg_for_dwarf(&mut self, reg: gimli::Register, val: u64) -> Option<()>; // None when register doesn't exist

    fn name_for_dwarf(reg: gimli::Register) -> Option<&'static str>; // None when register doesn't exist
    fn dwarf_for_name(name: &str) -> Option<gimli::Register>; // None when register doesn't exist
}
```

I am not quite sure how to handle different reg sizes on different architectures and within a single architecture though.

nbaksalyar:
This is a prerequisite for #96. I'd like to give it a try.

Stupremee:
Probably we can represent the different register sizes using an Enum with a u64, u32, ... variants and with a bunch of helper methods and `TryInto` impls it should be easy to get the raw values out of the Enum.

nbaksalyar:
Closed by #98

PR #98, opened 2020-08-29
Add API to read & write registers for Linux

nbaksalyar:
Implements #87.

- [x] Add trait definition.
- [x] Implement register functions for the Linux target.
- [x] Refactor the repl example to use the new API.
- [x] Refactor the gui example to use the new API.
- [x] Clean up the code and fix CI builds.

### DIFF
--- a/examples/repl.rs
+++ b/examples/repl.rs
@@ -512,7 +512,7 @@ mod example {
                         .function_debuginfo()
                         .ok_or_else(|| "No dwarf debuginfo for function".to_owned())?;
 
-                    let mut eval_ctx = X86_64EvalContext {
+                    let mut eval_ctx = EvalContext {
                         frame_base: None,
                         regs: Box::new(regs),
                     };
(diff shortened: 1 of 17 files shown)


examples/repl.rs
```diff
@@ -658,39 +658,9 @@ mod example {
...
-            };
-            match ty {
-                gimli::ValueType::Generic => gimli::Value::Generic(val),
-                gimli::ValueType::U64 => gimli::Value::U64(val),
-                _ => unimplemented!(),
-            }
-        }
-    }
-
     struct X86_64EvalContext {
```
bjorn3, line 661:
```suggestion
    struct EvalContext {
```
This is no longer x86_64 specific.

src/target/linux.rs
```diff
@@ -71,6 +91,9 @@ impl Thread for LinuxThread {
     }
 }
 
+/// Type alias to use for boxed threads.
+type BoxedLinuxThread = Box<dyn Thread<RegistersX86_64, ThreadId = i32>>;
```
bjorn3, line 95: Should get a `#[cfg]` as the register type depends on the target.

bjorn3 (changes requested):

nbaksalyar:
@bjorn3 should be fixed now, thanks!

bjorn3 (approved):

PR #101, opened 2020-08-31
Add a link to REPL demo recording to readme

nbaksalyar:


bjorn3:
Is the recording claimed? Otherwise it will ve deleted in a few days.

nbaksalyar:
It should be, I've signed up for an account and made the recording public.

PR #106, opened 2020-09-05
Add syntax highlighting and completion to repl

bjorn3:
image

### DIFF
--- a/examples/repl.rs
+++ b/examples/repl.rs
@@ -10,7 +10,11 @@ fn main() {
 
 #[cfg(target_os = "linux")]
 mod example {
-    use std::path::{Path, PathBuf};
+    use std::{
+        cell::Cell,
+        path::{Path, PathBuf},
+        rc::Rc,
+    };
 
     use headcrab::{
         symbol::{DisassemblySource, RelocatedDwarf},
@@ -44,6 +48,8 @@ mod example {
         Locals: String,
         /// Print this help
         Help|h: String,
+        /// Exit
+        Exit|quit|q: String,
     });
 
     type ReplHelper = repl_tools::MakeHelper<ReplCommand>;
@@ -83,13 +89,15 @@ mod example {
     }
 
     pub fn main() {
+        let color = Rc::new(Cell::new(true));
+
         let mut rl = rustyline::Editor::<ReplHelper>::with_config(
             rustyline::Config::builder()
                 .auto_add_history(true)
                 .completion_type(CompletionType::List)
                 .build(),
         );
-        rl.set_helper(Some(ReplHelper::default()));
+        rl.set_helper(Some(ReplHelper::new(color.clone())));
 
         let mut context = Context {
             remote: None,
@@ -111,6 +119,10 @@ mod example {
                         "Found flag -ex without argument".to_string()
                     }
                 }
+                "--no-color" => {
+                    color.set(false);
+                    continue;
+                }
                 _ if arg.starts_with("-") => {
                     format!("Found argument '{}' which wasn't expected", arg)
                 }
@@ -130,7 +142,8 @@ mod example {
         {} [OPTIONS] executable-file
 
     OPTIONS:
-        -ex <COMMAND>           Run command on startup",
+        -ex <COMMAND>           Run command on startup
+        --no-color              Disable colors",
                 err, repl_name
             );
             std::process::exit(1);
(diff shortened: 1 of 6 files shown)


bjorn3:
Would it make sense to move `repl_tools` to a separate repo, as it can be useful for other projects too?

examples/repl.rs
```diff
@@ -142,71 +209,101 @@ mod example {
...
+                context.set_remote(remote);
             }
-            Some("detach") => {
+            ReplCommand::Attach(pid) => {
+                let pid = nix::unistd::Pid::from_raw(pid.parse()?);
+                println!("Attaching to process {}", pid);
+                let (remote, status) = LinuxTarget::attach(
+                    pid,
+                    AttachOptions {
+                        kill_on_exit: false,
```
blitzerr, line 285: While we are at it, it might be good to move command implementation code, longer than a line into functions of their own ?

examples/repl.rs
```diff
@@ -142,71 +209,101 @@ mod example {
...
-                println!("\x1b[1mhelp|h\x1b[0m -- Print this help");
-                println!("\x1b[1mexit|quit|q\x1b[0m -- Exit");
+    fn run_command(
+        context: &mut Context,
+        color: &Cell<bool>,
+        command: &str,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        if command == "" {
+            return Ok(());
+        } else if command == "_patch_breakpoint_function" {
```
blitzerr, line 230: Should we put it in a method of its own ?

blitzerr:
Good one @bjorn3 !

examples/repl.rs
```diff
@@ -129,10 +192,14 @@ mod example {
                         println!("Exit");
                         return;
                     }
-                    match run_command(&mut context, &command) {
+                    match run_command(&mut context, &color, &command) {
                         Ok(()) => {}
                         Err(err) => {
-                            println!("\x1b[91mError: {}\x1b[0m", err);
+                            if color.get() {
+                                println!("\x1b[91mError: {}\x1b[0m", err);
```
nbaksalyar, line 199: Maybe use a dev-dependency like [`ansi_term`](https://crates.io/crates/ansi_term) to specify colours?

src/target/linux.rs
```diff
@@ -96,9 +97,10 @@ impl LinuxTarget {
 
     /// Launches a new debuggee process
     pub fn launch(
-        path: &str,
+        path: &Path,
```
nbaksalyar, line 100: We can use `impl AsRef<Path>` here instead (just as in the Rust's stdlib, e.g. in [`File::open`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.open)) - then we can pass the path as a string.

repl_tools/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    hint::Hinter,
+    validate::Validator,
+    Helper,
+};
+
+#[doc(hidden)]
+pub use rustyline as __rustyline;
+
+pub trait HighlightAndComplete: Sized {
+    fn from_str(line: &str) -> Result<Self, Box<dyn std::error::Error>>;
```
nbaksalyar, line 21: I think this might be a good place to start introducing concrete error types. I'd suggest we try using [`anyhow`](https://docs.rs/anyhow/), but no preference here (and this can be done at a later time, anyway).

nbaksalyar (commented):
This looks great, thanks! I left a few comments.

Re publishing `repl_tools` - that's an excellent idea. What's your preference here @bjorn3 - do you want to create a repository under the `headcrab-rs` organisation or have it under your username? Or would it be easier to just have it in-tree in this repo for now?

repl_tools/src/lib.rs
bjorn3: `anyhow` is for programs, not libraries. It doesn't really have a way to match on the contained error.

examples/repl.rs
bjorn3: `ansi_term` doesn't have a method for the bright variants of colors: https://github.com/ogham/rust-ansi-term/issues/66

bjorn3:
I think `repl_tools` can stay in tree for now. There may be multiple breaking changes in the near future. Once it becomes more stable, it can move to a separate repo as part of headcrab-rs.

nbaksalyar (approved):
Thank you!

Issue #100, opened 2020-08-30
Implement source mapping


nbaksalyar:
- Read source file locations from DWARF.
- Map addresses to lines in the source code (addr2line should be used for this).
- Read the source files and display the source code along with the line numbers at the current debugger location.

blitzerr:
I can work on this

blitzerr:
@nbaksalyar If you can add some mentoring notes to this, I would appreciate it.

bjorn3:
There is already a function to read the source line called [`Dwarf::source_snippet`](https://github.com/headcrab-rs/headcrab/blob/ab758ff2f5b7d719dfc4c87065d1d76ce720b9aa/src/symbol/source.rs#L71-L79). It only shows a single line though, so you may want to copy it to the `repl` example as new command and adapt it to show more lines, line numbers and maybe even syntax highlighting.

You could also use https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L63-L117 as starting point for this. It also depends on 
https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/syntax_highlight.rs for the syntax highlighting. It has line numbers, syntax highlighting and it dims all lines except for the current one. It also shows the current column with a `^`.

To also be able to view source for sysroot crates like libstd, you would also need to map files starting with `/rustc/` followed by a hash to the local path of the sysroot crates when the `rust-src` rustup component is installed. This could for example be done at https://github.com/headcrab-rs/headcrab/blob/ab758ff2f5b7d719dfc4c87065d1d76ce720b9aa/src/symbol/source.rs#L61-L64. See for example

https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L48-L61 and https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L71-L91

Due to changes in the sysroot source layout you would need to change https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L53 to I think `/lib/rustlib/library`.

blitzerr:
[...]

PR #107, opened 2020-09-05
[WIP] Implement source mapping

blitzerr:
Fixes: #100 

change summary
--------------

(+) Print source lines when pausing on a breakpoint.
(+) New command `list` or `l`
(+) Some code refactoring.

TODO:
-------
Waiting for @bjorn3 to merge [their PR](https://github.com/headcrab-rs/headcrab/pull/106) so that we can use it syntax highlight the source code we emit in this patch

currently if you debug your program with lldb, and say your program stops
at the set breakpoint, lldb will print the source line and not just filename
and line number. It might be more clear with an example, so here goes one:

```
rust-lldb tests/testees/hello
breakpoint set --file hello.rs --line 22
r

(lldb) r
Process 58912 launched: '/workspaces/headcrab/tests/testees/hello' (x86_64)
Process 58912 stopped
* thread #1, name = 'hello', stop reason = breakpoint 2.1
    frame #0: 0x0000555555559301 hello`hello::main::hd078db076938ab99 at hello.rs:22
   19           let mut temp = 100usize;
   20           black_box(&mut temp);
   21       }
-> 22       black_box(reg_var);
   23       breakpoint();
   24       black_box(reg_var);
   25       println!("{} {}", STATICVAR, var);
```

Before this PR, the behavior of headcrab for a similar steps of execution would
have been:

```
(headcrab) cargo run --example repl
(headcrab) exec tests/testees/hello
(headcrab) _patch_breakpoint_function
(headcrab) cont
(headcrab)
```
So, you can see, after cont(inuing) to the next breakpoint, headcrab is silent. You have to give
the command `bt` to print the backtrace.

```
(headcrab) bt
0000555555559295 core::core_arch::x86::sse2::_mm_pause /usr/local/rustup/toolchains/1.45.2-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/../stdarch/crates/core_arch/src/x86/sse2.rs:25
                 breakpoint /workspaces/headcrab/tests/testees/hello.rs:7
000055555555930b hello::main /workspaces/headcrab/tests/testees/hello.rs:24
0000555555559246 std::rt::lang_start::{{closure}} /usr/local/rustup/toolchains/1.45.2-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:67
0000555555562c68 std::panicking::try /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/panicking.rs:275
                 std::panic::catch_unwind /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/panic.rs:394
                 std::rt::lang_start_internal /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/rt.rs:51
```

After the change, the output will be something like this:

```
(headcrab) exec tests/testees/hello
Starting program: tests/testees/hello
Stopped(Pid(75698), SIGTRAP)
(headcrab) _patch_breakpoint_function
(headcrab) cont
Stopped(Pid(75698), SIGTRAP)
0000555555559295 core::core_arch::x86::sse2::_mm_pause /usr/local/rustup/toolchains/1.45.2-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/../stdarch/crates/core_arch/src/x86/sse2.rs:25
/workspaces/headcrab/tests/testees/hello.rs:7:14
    4 #[inline(never)]
    5 fn breakpoint() {
    6     // This will be patched by the debugger to be a breakpoint
>   7     unsafe { core::arch::x86_64::_mm_pause(); }
    8 }
    9
   10 #[inline(never)]
(headcrab)
```
So, as you can see, now when we pause at the break-point, we also print the source code around it.


blitzerr:
@bjorn3 @nbaksalyar  r?

examples/repl.rs
```diff
@@ -198,7 +210,12 @@ mod example {
             }
             Some("kill") => println!("{:?}", context.remote()?.kill()?),
             Some("si") | Some("stepi") => println!("{:?}", context.remote()?.step()?),
-            Some("cont") | Some("continue") => println!("{:?}", context.remote()?.unpause()?),
+            Some("cont") | Some("continue") => {
+                println!("{:?}", context.remote()?.unpause()?);
+                // When we hit the next breakpoint, we also want to display the source code
+                // as does lldb and gdb.
+                print_source_for_top_of_stack_symbol(context, 3)?;
```
bjorn3, line 217: This should also happen on `stepi`, `exec` and `attach` I think. Basically every command that causes execution to pause.

examples/repl.rs
```diff
@@ -431,6 +425,143 @@ mod example {
...
+    /// Get the call_stack from the context and then tries to display the
+    /// source for the top call in the stack. Becasue the first frame is usually
+    /// sse2.rs, we just display the file and line but not the source and we skip
+    /// over to the next frame. For the next frame, we will display the source code.
+    /// An example view is shown below:
+    ///
+    /// It marks the line with the berakpoint with a '>' character and shows some lines
+    /// of context above and below it.
+    ///
+    /// 0000555555559295 core::core_arch::x86::sse2::_mm_pause /../rustup/toolchains/1.45.2-x86_64-unknown-linux-gnu/../stdarch/crates/core_arch/src/x86/sse2.rs:25
```
bjorn3, line 473: Please put this in a ```` ```plain ```` code block.

examples/repl.rs
```diff
@@ -431,6 +425,143 @@ mod example {
...
+                            name,
+                            format!("{}:{}", file, line)
+                        );
+                    } else {
+                        print_file_lines_with_context(file, line, column, context_lines)?;
+                    }
+
+                    first_frame = false;
+                }
+                Ok(first_frame)
```
bjorn3, line 526:
```suggestion
                Ok(())
```

examples/repl.rs
```diff
@@ -431,6 +425,143 @@ mod example {
...
+                        })
+                        .unwrap_or_default();
+
+                    if first_frame {
+                        println!(
+                            "{:016x} {} {}",
+                            top_of_stack,
+                            name,
+                            format!("{}:{}", file, line)
+                        );
```
bjorn3, line 519: This is wrong when you stepped out of `_mm_pause`. Maybe just specifically skip any function with `_mm_pause` in the name. (And add a FIXME to remove it once `_patch_breakpoint_function` is gone)

examples/repl.rs
```diff
@@ -431,6 +425,143 @@ mod example {
...
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let file = File::open(&filename)?;
+        let reader = BufReader::new(file);
+
+        let mut i = 0;
+        let start = std::cmp::max(0, line_no - context_lines);
+        let end = line_no + context_lines;
+
+        println!("{}:{}:{}", filename, line_no, column);
+        for line in reader.lines() {
```
bjorn3, line 545:
```suggestion
        for (i, line) in reader.lines().enumerate().skip(start) {
```

examples/repl.rs
```diff
@@ -431,6 +425,143 @@ mod example {
...
+        filename: &str,
+        line_no: u32,
+        column: u32,
+        context_lines: u32,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let file = File::open(&filename)?;
+        let reader = BufReader::new(file);
+
+        let mut i = 0;
+        let start = std::cmp::max(0, line_no - context_lines);
```
bjorn3, line 541: This will be a panic on underflow in debug mode. You can use `saturating_sub` instead I think.

bjorn3 (changes requested):
Nice!

examples/repl.rs
blitzerr: makes sense. Done !

examples/repl.rs
blitzerr: done !

examples/repl.rs
blitzerr: I have added it to stepi. But does it make sense to add it to exec. I don't see lldb doing that.

examples/repl.rs
blitzerr: Looks like this is a nightly only feature : https://doc.rust-lang.org/std/intrinsics/fn.saturating_sub.html ?

examples/repl.rs
blitzerr: Ohh ! This is neat. :)

examples/repl.rs
blitzerr: I have made changes to take care of underflow. Please take a look and let me know if there is a better rusty way of doing it :)

examples/repl.rs
```diff
@@ -431,6 +431,144 @@ mod example {
...
+                                )
+                            })
+                            .unwrap_or_default();
+
+                        print_file_lines_with_context(
+                            file,
+                            line as usize,
+                            column,
+                            context_lines as usize,
+                        )?;
```
bjorn3, line 529: Please ads a return here to avoid printing multiple source lines in case of inlining.

examples/repl.rs
bjorn3: For exec I guess omitting it makes sense.

examples/repl.rs
```diff
@@ -163,6 +175,7 @@ mod example {
                 println!("\x1b[1mregisters|regs\x1b[0m read -- List registers and their content for the current stack frame");
                 println!("\x1b[1mbacktrace|bt\x1b[0m -- Print backtrace of stack frames");
                 println!("\x1b[1mdisassemble|dis\x1b[0m -- Print the disassembled source");
+                println!("\x1b[1mlist|l\x1b[0m -- Print the source code exetution point.");
```
nbaksalyar, line 178: typo:
```suggestion
                println!("\x1b[1mlist|l\x1b[0m -- Print the source code execution point.");
```

examples/repl.rs
```diff
@@ -21,6 +24,15 @@ mod example {
         disassembler: DisassemblySource,
     }
 
+    /// The subcommands that are acceptable for backtrace.
+    enum BtType {
```
nbaksalyar, line 28: Nit - I think having a more explicit name would convey the intent better:
```suggestion
    enum BacktraceType {
```

examples/repl.rs
```diff
@@ -207,37 +230,13 @@ mod example {
...
-                    .collect(),
-                    Some("naive") => headcrab::symbol::unwind::naive_unwinder(
-                        context.debuginfo(),
-                        &stack[..],
-                        regs.rip as usize,
-                    )
-                    .collect(),
+                let bt_type = match parts.next() {
+                    Some("fp") | None => BtType::FramePtr,
+                    Some("naive") => BtType::Naive,
```
nbaksalyar, line 235: Maybe move this code to `impl BtType`? Then we can have it all in one place - e.g.:
```rust
impl BtType {
    fn from_str(value: Option<&str>) {
        match value {
            Some("fp") | None => BtType::FramePtr,
            Some("naive") => BtType::default(),
        }
    }
}
```

examples/repl.rs
```diff
@@ -431,6 +431,144 @@ mod example {
...
+                context.debuginfo(),
+                &stack[..],
+                regs.rip as usize,
+            )
+            .collect(),
+        };
+        Ok(call_stack)
+    }
+
+    /// Get the call_stack from the context and then tries to display the
```
nbaksalyar, line 470: typo:
```suggestion
    /// Gets the call_stack from the context and then tries to display the
```

examples/repl.rs
```diff
@@ -431,6 +431,144 @@ mod example {
...
+    /// source for the top call in the stack. Becasue the first frame is usually
+    /// sse2.rs, we just display the file and line but not the source and we skip
+    /// over to the next frame. For the next frame, we will display the source code.
+    /// An example view is shown below:
```
nbaksalyar, line 474: Brilliant docs, thanks!

examples/repl.rs
```diff
@@ -431,6 +431,144 @@ mod example {
...
+                &stack[..],
+                regs.rip as usize,
+            )
+            .collect(),
+        };
+        Ok(call_stack)
+    }
+
+    /// Get the call_stack from the context and then tries to display the
+    /// source for the top call in the stack. Becasue the first frame is usually
```
nbaksalyar, line 471: typo:
```suggestion
    /// source for the top call in the stack. Because the first frame is usually
```

nbaksalyar (changes requested):
Nice one, thank you!
I left a few minor suggestions and noticed a few typos.

examples/repl.rs
bjorn3: This can be done after #106 is merged. `HighlightAndComplete` has a `from_str` method. Implementing it would also give autocomplete and syntax highlighting.

examples/repl.rs
blitzerr: I was thinking about it.Thanks for pointing this out.

examples/repl.rs
blitzerr: Makes sense.

examples/repl.rs
blitzerr: So I would just leave a comment for now here and make the refactoring after #106 is merged ?

bjorn3: #106 was merged 4 hours ago.

blitzerr: :)

Just saw it. So I will get on with it

blitzerr:
@nbaksalyar @bjorn3 the build is failing on 
test target::linux::writemem::tests::write_protected_memory ... FAILED

This patch should not impact that, right ?

bjorn3:
Strange. What is the failure message?

blitzerr:
> Strange.

It didn't fail earlier but after rebasing ..

blitzerr:
@bjorn3 and @nbaksalyar You comments are addressed and rebased with the latest.

examples/repl.rs
```diff
@@ -20,6 +22,8 @@ mod example {
     use repl_tools::HighlightAndComplete;
     use rustyline::CompletionType;
 
+    pub type CrabResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;
```
bjorn3, line 25: Using `Result` here should work, right?

examples/repl.rs
```diff
@@ -58,6 +64,25 @@ mod example {
         disassembler: DisassemblySource,
     }
 
+    /// The subcommands that are acceptable for backtrace.
+    enum BacktraceType {
```
bjorn3, line 68: Please implement `HighlightAndComplete` for `BacktraceType` and at line 46 (`Backtrace|bt: String`) instead of `String`.

examples/repl.rs
```diff
@@ -58,6 +64,25 @@ mod example {
...
+        // uses naive_unwinder.
+        Naive,
+    }
+
+    impl BacktraceType {
+        fn from_str(value: &str) -> CrabResult<Self> {
+            match value {
+                "fp" | "" => Ok(BacktraceType::FramePtr),
+                "naive" => Ok(BacktraceType::Naive),
+                &_ => Err(format!("Unrecognized backtrace type {}. Supported ones are 'fp' and 'naive'. Please consider using one of them.", value).into()),
```
bjorn3, line 81:
```suggestion
                _ => Err(format!("Unrecognized backtrace type {}. Supported ones are 'fp' and 'naive'. Please consider using one of them.", value).into()),
```

examples/repl.rs
```diff
@@ -256,8 +281,16 @@ mod example {
...
-            ReplCommand::Stepi(()) => println!("{:?}", context.remote()?.step()?),
-            ReplCommand::Continue(()) => println!("{:?}", context.remote()?.unpause()?),
+            ReplCommand::Stepi(()) => {
+                println!("{:?}", context.remote()?.step()?);
+                return print_source_for_top_of_stack_symbol(context, 3);
+            }
+            ReplCommand::Continue(()) => {
+                println!("{:?}", context.remote()?.unpause()?);
+                // When we hit the next breakpoint, we also want to display the source code
+                // as does lldb and gdb.
```
bjorn3, line 291: This sentence doesn't quite sound grammatically correct to me.

nbaksalyar:
> the build is failing on
> test target::linux::writemem::tests::write_protected_memory ... FAILED

Hmm, probably an intermittent error related to #95 - I'll try to run separate fuzz-like tests with that.

examples/repl.rs
```diff
@@ -503,6 +510,153 @@ mod example {
...
+                    let name = frame
+                        .function
+                        .as_ref()
+                        .map(|f| Ok(f.demangle()?.into_owned()))
+                        .transpose()
+                        .map_err(|err: gimli::Error| err)?
+                        .unwrap_or_else(|| "<unknown>".to_string());
+
+                    // TODO: currently we are skipping over mm_pause but once we have real
+                    // breakpoint support and _patch_breakpoint_function gets removed, this
```
bjorn3, line 588:
```suggestion
                    // TODO: currently we are skipping over `_mm_pause` but once we have real
                    // breakpoint support and `_patch_breakpoint_function` gets removed, this
```

nit

examples/repl.rs
```diff
@@ -503,6 +510,153 @@ mod example {
...
+        };
+        let end = key + context_lines;
+
+        println!("{}:{}:{}", filename, line_no, column);
+
+        for (i, line) in reader.lines().enumerate().skip(start) {
+            {
+                //let l = line.unwrap().clone();
+                //println!("{}", String::highlight(l.as_str()));
+            }
```
bjorn3, line 641:
```suggestion
```

examples/repl.rs
```diff
@@ -599,7 +753,6 @@ mod example {
             local.name()?.unwrap_or("<no name>"),
             value
         );
-
```
bjorn3, line 602: Please keep this.

examples/repl.rs
```diff
@@ -503,6 +510,153 @@ mod example {
...
+            0
+        };
+        let end = key + context_lines;
+
+        println!("{}:{}:{}", filename, line_no, column);
+
+        for (i, line) in reader.lines().enumerate().skip(start) {
+            {
+                //let l = line.unwrap().clone();
+                //println!("{}", String::highlight(l.as_str()));
```
nbaksalyar, line 640: I think this is commented-out code and it can be removed?

bjorn3 (changes requested):

nbaksalyar (commented):
LGTM, thank you! I've just one minor comment about commented-out code

examples/repl.rs
blitzerr: you mena not the fully qualified `std::result::Result` but just Result, right ?

examples/repl.rs
blitzerr: @bjorn3 Can you elaborate more on this ?

blitzerr: @bjorn3 Please pardon my pure rust skills. Let me be explicit and ask, you want the line 46 to change from

```rust
Backtrace|bt: String,
```
to 

```rust
Backtrace|bt: (Backtrace|bt: String),
```

bjorn3: No, change it to `Backtrace|bt: BacktraceType,`. That requires implementing `HighlightAndComplete` for `BacktraceType`.

blitzerr: done!

examples/repl.rs
```diff
@@ -58,6 +61,67 @@ mod example {
...
+    impl Default for BacktraceType {
+        fn default() -> Self {
+            BacktraceType::FramePtr
+        }
+    }
+
+    impl HighlightAndComplete for BacktraceType {
+        type Error = VoidError;
+        fn from_str(line: &str) -> Result<Self, Self::Error> {
+            Ok(BacktraceType::from_str(line.trim()).unwrap_or_default())
```
bjorn3, line 94: Could you inline this call and actually return the error when there is any? You could use `type Error = String` above.

src/symbol/source.rs
```diff
@@ -78,3 +81,155 @@ impl super::Dwarf {
             .to_string())
     }
 }
+
+pub type CrabResult<T> = Result<T, Box<dyn std::error::Error>>;
+
+#[derive(Debug)]
+/// A line in a source code is represented as a line number and the string.
```
bjorn3, line 88:
```suggestion
/// A line in a source code is represented as a line number and the string.
#[derive(Debug)]
```

Doc comments are often written before the attributes of a type.

src/symbol/source.rs
```diff
@@ -78,3 +81,155 @@ impl super::Dwarf {
...
+            file_path: file_path.to_string(),
+            fn_name,
+            lines,
+            key_line_idx: lines_as_context,
+            key_column_idx: column - 1,
+        })
+    }
+}
+pub mod pretty {
+    use super::{Snippet, SourceLine};
```
bjorn3, line 161: Maybe add a comment that this is adapted from https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/syntax_highlight.rs and https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L103-L113?

blitzerr: Good point. It somehow slipped my mind.

examples/repl.rs
```diff
@@ -70,14 +70,23 @@ mod example {
         Naive,
     }
 
+    #[derive(Debug)]
+    struct BacktraceTypeError(String);
+
+    impl std::fmt::Display for BacktraceTypeError {
+        fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+            write!(f, "Unrecognized backtrace type {}. Supported ones are 'fp' and 'naive'. Please consider using one of them.", self.0.trim())
```
bjorn3, line 78:
```suggestion
            write!(f, "Unrecognized backtrace type `{}`. Supported ones are 'fp' and 'naive'. Please consider using one of them.", self.0.trim())
```

nit

examples/repl.rs
```diff
@@ -70,14 +70,23 @@ mod example {
...
+        }
+    }
+    impl std::error::Error for BacktraceTypeError {}
+
     impl BacktraceType {
-        fn from_str(value: &str) -> CrabResult<Self> {
+        #[inline]
+        fn from_str(value: &str) -> Result<Self, BacktraceTypeError> {
```
bjorn3, line 85: I actually meant putting the body of this function in the `from_str` of the `HighlightAndComplete` impl and removing this function. Not marking this as `#[inline]`.

blitzerr: But `from_str` goes with the BacktraceType implementation and not ` impl HighlightAndComplete for BacktraceType`. So, this might be the way to go

bjorn3: As discussed on PM this can be resolved later.

bjorn3 (approved):

Cargo.toml
```diff
@@ -17,6 +17,7 @@ object = "0.20"
 gimli = "0.22.0"
 capstone = "0.7.0"
 addr2line = "0.13.0"
+syntect = "4.4.0"
```
nbaksalyar, line 20: For now, this is used only by `examples/repl.rs`. I think it'd be good to move this to the `[dev-dependencies]` section below. However, the `Snippet` helper is in the main source tree and it might be helpful for other use cases too, so we can't easily make this a dev-dependency.

Maybe we can use a feature flag instead? So `mod pretty` can be compiled conditionally with `#[cfg(feature = "syntax-hughlighting")]` and the dependency can be included as `syntect = { version = "4.4.0", optional = true }`. What do you think?

nbaksalyar (commented):
Thank you, great work! :+1: 
I've got one question about a new dependency which we can consider moving behind a feature-flag.

Cargo.toml
bjorn3: `syntect` is used in `headcrab::symbol::source::pretty`, which doesn't have access to dev dependencies. I agree that a feature flag would be nice.

blitzerr: @nbaksalyar I have added the feature gates in the latest revision. But now we need to run the repl example as
`cargo run --example repl --features syntax-highlighting`. Is that okay ?

blitzerr: Also, if the feature is not enabled, the examples should still work

src/symbol.rs
```diff
@@ -24,7 +24,7 @@ mod sym;
 
 pub use frame::{Frame, FrameIter, Local, LocalValue};
 pub use relocate::RelocatedDwarf;
-pub use source::DisassemblySource;
+pub use source::{pretty, CrabResult, DisassemblySource, Snippet};
```
bjorn3, line 27:
```suggestion
pub use source::{CrabResult, DisassemblySource, Snippet};
```

No need to re-export `pretty`.

src/symbol/source.rs
```diff
@@ -78,3 +81,159 @@ impl super::Dwarf {
...
+            key_line_idx: lines_as_context,
+            key_column_idx: column - 1,
+        })
+    }
+}
+
+/// This module has methods to syntax highlighting of the code snippet.
+/// This is an adaptation of [this](https://github.com/bjorn3/pretty_backtrace/blob/c8a19b94f369a2edd9d58b07eb4310d2e0a2c991/src/display_frame.rs#L103-L113).
+#[cfg(feature = "syntax-highlighting")]
+pub mod pretty {
```
bjorn3, line 164:
```suggestion
mod pretty {
```

This can be private. Module publicness doesn't affect if methods are public.

src/symbol.rs
blitzerr: I see.

src/symbol/source.rs
blitzerr: ahh ! I learned that now :) . WIll make the change.

blitzerr:
@nbaksalyar r?

bjorn3:
Please run `cargo fmt`.

Cargo.toml
```diff
@@ -44,4 +45,9 @@ repl_tools = { path = "./repl_tools" }
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 headcrab_inject = { path = "./headcrab_inject" }
 
+[features]
+syntax-highlighting = ["syntect"]
+
 [workspace]
+
+
```
bjorn3, line 53: Please remove these extra newlines.

blitzerr: done

nbaksalyar:
Hmm, the build seems to be failing:

```
error[E0252]: the name `HighlightAndComplete` is defined multiple times
  --> examples/repl.rs:30:9
   |
29 |     use repl_tools::HighlightAndComplete;
   |         -------------------------------- previous import of the trait `HighlightAndComplete` here
30 |     use repl_tools::HighlightAndComplete;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `HighlightAndComplete` reimported here
   |
   = note: `HighlightAndComplete` must be defined only once in the type namespace of this module
```

https://travis-ci.com/github/headcrab-rs/headcrab/jobs/383054168#L426-L444

nbaksalyar (approved):

bjorn3:
Thanks @blitzerr!

blitzerr:
My first PR merged :D

Issue #102, opened 2020-09-03
Use an `Arch` enum and feature flags instead of cfg(target_arch)


nbaksalyar:
Instead of using `#[cfg(target_arch = "x86_64")]` for architecture-specific implementations, introduce corresponding feature flags - e.g.: `#[cfg(feature = "x86_64")]`.

In addition to that, introduce an enum to be used for type arguments:

```rust
enum Arch { x86_64 }
```

This is required for the possibility of future multi-arch builds of Headcrab - for example, in case if we'd want to work with a core dump generated on an ARM-based machine (or connect to that machine remotely using the gdb-serial target).

PR #112, opened 2020-09-11
refactor(target_arch): pass arch via feature flags instead

kakoc:
Related to: https://github.com/headcrab-rs/headcrab/issues/102  
It's not clear to me: should the new `Arch` enum be used anywhere now?

### DIFF
--- a/headcrab_inject/src/lib.rs
+++ b/headcrab_inject/src/lib.rs
@@ -1,5 +1,5 @@
 // FIXME make this work on other systems too.
-#![cfg(all(target_arch = "x86_64", target_os = "linux"))]
+#![cfg(all(feature = "x86_64", target_os = "linux"))]
 
 use std::{collections::HashMap, error::Error};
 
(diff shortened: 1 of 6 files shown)


src/target/linux.rs
```diff
@@ -24,17 +24,17 @@ pub use writemem::WriteMemory;
 
 lazy_static::lazy_static! {
     pub static ref PAGE_SIZE: usize = unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize };
-    #[cfg(target_arch="x86_64")]
+    #[cfg(feature="x86_64")]
```
bjorn3, line 27: Ptrace can only work with the local arch, so using `target_arch` makes more sense to me. Also enabling the x86_64 feature on other archs won't compile, as libc won't jave a u_debugreg field for the `libc::user` struct.

headcrab_inject/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 // FIXME make this work on other systems too.
-#![cfg(all(target_arch = "x86_64", target_os = "linux"))]
+#![cfg(all(feature = "x86_64", target_os = "linux"))]
```
bjorn3, line 2: `headcrab_inject` directly uses the rip and rsp fields of `libc::user_regs`, so it won't compile for other archs.

bjorn3:
`headcrab_inject` could be made to work with multiple archs, but that would likely require #87 first.

PR #113, opened 2020-09-19
Refactor local printing

bjorn3:
This fixes a panic when trying to print a struct. It shows `<struct>` for now until I implement real support for structs.

### DIFF
--- a/src/symbol.rs
+++ b/src/symbol.rs
@@ -22,7 +22,7 @@ mod relocate;
 mod source;
 mod sym;
 
-pub use frame::{Frame, FrameIter, Local, LocalValue};
+pub use frame::{Frame, FrameIter, Local, LocalValue, PrimitiveValue};
 pub use relocate::RelocatedDwarf;
 pub use source::{CrabResult, DisassemblySource, Snippet};
 
--- a/src/symbol.rs
+++ b/src/symbol.rs
@@ -1,6 +1,7 @@
 //! This module provides a naive implementation of symbolication for the time being.
 //! It should be expanded to support multiple data sources.
 
+#[macro_use]
 pub mod dwarf_utils;
 pub mod unwind;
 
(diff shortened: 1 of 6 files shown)


src/symbol.rs
```diff
@@ -1,6 +1,7 @@
 //! This module provides a naive implementation of symbolication for the time being.
 //! It should be expanded to support multiple data sources.
 
+#[macro_use]
```
nbaksalyar, line 4: I think we can follow the 2018 edition style here, so we can replace this with
```rust
use dwarf_utils::dwarf_attr;
```

src/symbol/dwarf_utils.rs
```diff
@@ -1,7 +1,94 @@
-use gimli::{DebuggingInformationEntry, Dwarf, Unit};
+use gimli::{DebuggingInformationEntry, Dwarf, Unit, UnitOffset, ValueType};
 
 use super::Reader;
 
+macro_rules! dwarf_attr {
+    (str($dwarf:ident,$unit:ident) $entry:ident.$name:ident || $missing:ident) => {
+        if let Some(attr) = $entry.attr(gimli::$name)? {
+            dwarf_attr_not_missing_action!($missing, $dwarf.attr_string(&$unit, attr.value())?)
```
nbaksalyar, line 8: Nitpick - I'd suggest using a more descriptive name, e.g.
```suggestion
            dwarf_attr_exists_action!($missing, $dwarf.attr_string(&$unit, attr.value())?)
```

nbaksalyar (commented):
It looks _so_ much better with `EvalContext`!
I left a couple of suggestions & nitpicks about style/naming.

src/symbol.rs
bjorn3: As far as I know that only works for extern crates and the unstable decl macro 2.0.

bjorn3:
When a review suggests only small changes can I merge after resolving them and having all tests pass or do you want to review again @nbaksalyar?

nbaksalyar (approved):
Sure! When there are no major changes after a review, please feel free to merge.

Thank you :+1:

PR #114, opened 2020-09-24
Add imgui based gui example

bjorn3:
image

It looks much better with a ttf font instead of the builtin imgui font, but I will leave that for a future PR.


nbaksalyar (approved):
Looks great, thanks! Would you also like to add the font files?

macOS CI run fails with `the remote end hung up unexpectedly` which seems like a problem on the Travis's end. My vote is for ignoring this temporarily - there will be more PRs dealing with missing features for macOS anyway.

bjorn3:
I will add a font tomorrow. The travis issue looks like a temporary network outage.

bjorn3:
image

New screenshot with FiraMono as font.

bjorn3:
```
---- target::linux::writemem::tests::write_memory_ptrace stdout ----

thread 'target::linux::writemem::tests::write_memory_ptrace' panicked at 'called `Result::unwrap()` on an `Err` value: Sys(EPERM)', src/target/linux/writemem.rs:292:91
```

I didn't change any of the relevant code, but there also doesn't seem to be a race-condition.

nbaksalyar:
This seems like a bug in the ptrace tests. I will create a separate issue to tackle it, and I restarted the CI - let's see if it shows up consitently.

bjorn3:
Happened again.

nbaksalyar:
And now it's passed. I created #116 and will investigate it as a separate issue.
Merging this PR now.

Thanks!

PR #108, opened 2020-09-06
Breakpoint support on Linux

Arthamys:
Here is a first version that supports setting and removing breakpoints, as  well as a single_step function.
I wanted to get your feedback on the interface, as well as some inputs on the edge cases that should be handled.

Bjorn mentioned in #55 that code should handle "faulting instructions at the place of the breakpoint" but I don't know what should be the particular reaction we should have in this situation. Should we not allow settings the breakopint?

PR #117, opened 2020-09-28
Accept an arbitrary Command as argument for launch

bjorn3:
This allows passing arguments and setting env vars for the debuggee

### DIFF
--- a/src/target/unix.rs
+++ b/src/target/unix.rs
@@ -1,10 +1,9 @@
 use nix::{
     sys::ptrace,
     sys::wait::{waitpid, WaitStatus},
-    unistd::{execv, fork, ForkResult, Pid},
+    unistd::Pid,
 };
-use std::ffi::CString;
-use std::process;
+use std::process::Command;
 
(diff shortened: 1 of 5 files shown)


src/target/unix.rs
```diff
@@ -41,40 +40,27 @@ pub trait UnixTarget {
...
-        }
-        ForkResult::Child => {
-            if let Err(err) = ptrace::traceme() {
-                println!("ptrace traceme failed: {:?}", err);
-                process::abort()
-            }
-
+    use std::os::unix::process::CommandExt;
+    unsafe {
+        cmd.pre_exec(|| {
```
nbaksalyar, line 47: That's really neat :)

nbaksalyar (approved):
LGTM, thanks! :+1: 

This PR conflicts with #108 though, so if that's OK, I'd prefer merging the larger breakpoints PR first, and then rebasing this one.

bjorn3:
It actually didn't conflict with #108. I rebased anyway.

Issue #116, opened 2020-09-27
Fix sporadic bug in the write_ptrace test


nbaksalyar:
It looks like a race condition that hasn't been observed in local environments yet.
The error occurs in the `write_memory_ptrace` test:

```
---- target::linux::writemem::tests::write_memory_ptrace stdout ----

thread 'target::linux::writemem::tests::write_memory_ptrace' panicked at 'called `Result::unwrap()` on an `Err` value: Sys(EPERM)', src/target/linux/writemem.rs:292:91

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

failures:

    target::linux::writemem::tests::write_memory_ptrace
```

A possible strategy to find the root case:

1. Replicate the Travis CI environment (same Docker env settings and capabilities).
2. Set a breakpoint on the `if let Err(Sys(EPERM)) = result { ... }` condition.
3. Run the test repeatedly until the breakpoint is hit, and observe the local state.

PR #118, opened 2020-09-30
remove _patch_breakpoint function

Arthamys:


bjorn3:
Could you also change https://github.com/headcrab-rs/headcrab/blob/e4eb3f3b69050c8c89aad972168b500d1c7e108c/tests/test_utils.rs#L47 to use the new breakpoint support and remove the line at https://github.com/headcrab-rs/headcrab/blob/e4eb3f3b69050c8c89aad972168b500d1c7e108c/tests/testees/hello.rs#L7

bjorn3:
You may need to change the amount of `single_step()` calls in some tests to adjust for the new instruction counts.

Arthamys:
> remove the line at
> 
> https://github.com/headcrab-rs/headcrab/blob/e4eb3f3b69050c8c89aad972168b500d1c7e108c/tests/testees/hello.rs#L7

This should probably be kept, otherwise the compiler will just optimize away the function.
Or we can replace it by some other meaningless computation if you'd rather

bjorn3:
You can keep it. I forgot that LLVM is smart enough to remove empty `#[inline(never)]` functions.

nbaksalyar (approved):
This looks good, thanks! Could you please rebase it on latest master so that it can be merged?

Arthamys:
The checks failed due to #116

bjorn3:
Thanks!

PR #122, opened 2020-10-11
Add error type wrapping all possible errors

DevinR528:
Just wondering if I'm moving in the right direction? Any naming suggestions/ideas would be helpful as I am still getting my head around everything.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,41 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CrabError {
+    #[error("Error occurred in headcrab: {0}")]
+    HeadCrab(String),
+
+    #[error("{0}")]
+    Dwarf(String),
+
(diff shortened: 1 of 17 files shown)


src/error.rs
```diff
@@ -0,0 +1,41 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CrabError {
```
bjorn3, line 4: I think there should be a separate error type for target and one for symbol.

DevinR528: Does there still need to be a top-level wrapper for `pub type CrabResult<T> = Result<T, ErrorType>;`?

DevinR528: Or should each mod (target and symbol) have their own `Result` type?

bjorn3: > Does there still need to be a top-level wrapper for pub type CrabResult<T> = Result<T, ErrorType>;? Something like

That could be useful.

PR #135, opened 2022-04-02
add a `munmap` method for `LinuxTarget`

pro465:
i found there are many `// FIXME`s about yet having to implement `munmap`ping the mapped files, so i thought that was a good place to start?

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -306,6 +306,22 @@ impl LinuxTarget {
         )
     }
 
+    /// unmap the memory.
+    pub unsafe fn munmap(
+        &self,
+        addr: *mut libc::c_void,
+        length: libc::size_t,
+    ) -> CrabResult<()> {
+        if libc::munmap(
+            addr,
+            length,
+        ) == -1 {
+             Err("mmunmap failed")
+        } else {
+             Ok(())
+        }
+    }
+
     pub fn memory_maps(&self) -> CrabResult<Vec<super::MemoryMap>> {
         Ok(procfs::process::Process::new(self.pid.as_raw())?
             .maps()?
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -314,6 +314,7 @@ impl LinuxTarget {
     ) -> CrabResult<libc::c_ulonglong> {
         self.syscall(libc::SYS_munmap as _, addr as _, length as _, 0, 0, 0, 0)
     }
+
     pub fn memory_maps(&self) -> CrabResult<Vec<super::MemoryMap>> {
         Ok(procfs::process::Process::new(self.pid.as_raw())?
             .maps()?
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -307,19 +307,16 @@ impl LinuxTarget {
     }
 
     /// unmap the memory.
-    pub unsafe fn munmap(
-        &self,
-        addr: *mut libc::c_void,
-        length: libc::size_t,
-    ) -> CrabResult<()> {
-        if libc::munmap(
-            addr,
-            length,
-        ) == -1 {
-             Err("mmunmap failed")
-        } else {
-             Ok(())
-        }
+    pub fn munmap(&self, addr: *mut libc::c_void, length: libc::size_t) -> CrabResult<libc::c_ulonglong> {
+        self.syscall(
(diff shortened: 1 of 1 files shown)


src/target/linux.rs
```diff
@@ -306,6 +306,22 @@ impl LinuxTarget {
         )
     }
 
+    /// unmap the memory.
+    pub unsafe fn munmap(
+        &self,
+        addr: *mut libc::c_void,
+        length: libc::size_t,
+    ) -> CrabResult<()> {
+        if libc::munmap(
```
bjorn3, line 315: This would unmap the memory inside the debugger process, not the debugged process. You need to use `self.syscall(libc::SYS_unmap as _, addr as _, length as _)`

pro465: uh...i guess you are right since i dont know much about this 


but, what about the additional arguments? i initially wanted to call `self.syscall`, but could not figure out what should i pass to the additional arguments. and i was afraid to pass any value, since the code for self.syscall seems to be modifying registers.

bjorn3: You can pass in 0. `self.syscall` restores all registers to their original value after performing the syscall: https://github.com/headcrab-rs/headcrab/blob/8e9424a8a9d650a6a0242b0cbbb5f8e7388f9037/src/target/linux.rs#L283

pro465: ah yes, thanks for the reassurance and directions!

nbaksalyar:
@pro465, do you want to also add a small test for it? Basically, it can go like

1. Add a memory mapping (call `target.mmap()` with an address and a length)
2. Check that the memory mapping has been added (call `target.memory_maps()` and verify that the expected address is in the results list)
3. Remove the memory mapping (call `target.munmap()` with the same address & length)
4. Check that it's been removed (call `target.memory_maps()` again and verify that the memory map is no longer in the list)

We don't have tests for manipulating memory maps yet so it would be great to have this. But this can go in a separate PR or added at a later time by me or other contributors.

pro465:
i'd favor the first option, as i m in favor of small changes for now.

i will add tests later in a separate PR.

src/target/linux.rs
```diff
@@ -306,6 +306,14 @@ impl LinuxTarget {
...
     }
 
+    /// unmap the memory.
+    pub fn munmap(
+        &self,
+        addr: *mut libc::c_void,
+        length: libc::size_t,
+    ) -> CrabResult<libc::c_ulonglong> {
+        self.syscall(libc::SYS_munmap as _, addr as _, length as _, 0, 0, 0, 0)
+    }
```
nbaksalyar, line 316: Very minor nitpick – could you please add a missing newline after the closing curly bracket?

```suggestion
    }

```

nbaksalyar (approved):
Thank you!

pro465:
welcome!

do you want me to add tests in this PR?

pro465:
so, let me outline the test:
```
fn test_mmap() {
    let target = LinuxTarget::new(/* ??? */);
    target.mmap(0 as *mut _, 10);
    let mmapped_mem = /* ??? */;
    assert!(target.memory_maps().contains(&mmapped_mem));
    target.munmap(mmapped_mem);
    assert!(!target.memory_maps().contains(&mmapped_mem));
}
```

pro465:
@nbaksalyar hello?

nbaksalyar:
Hey, I've been just waiting to see if there will be any other review comments.
LGTM! 👍 

And yeah, the test is looking good - as for this part

>     let target = LinuxTarget::new(/* ??? */);

you can take the same approach as with the other tests which use [a set of well-known programs](
https://github.com/headcrab-rs/headcrab/tree/master/tests/testees). You can even use one of the existing 'testees', e.g. [hello.rs](https://github.com/headcrab-rs/headcrab/blob/04cb3104dd453273d2e4c5bd528f7ba23d5099c6/tests/testees/hello.rs).

pro465:
@nbaksalyar thank you again! btw, should i just include the test in tests/syscall.rs, or should i create a new test file?

nbaksalyar:
@pro465, yep, tests/syscall.rs should be the ideal module for this. :)

pro465:
thank you for the info!, see ya in the next PR

PR #136, opened 2022-04-03
add m[un]map tests

pro465:
I'm super nervous about whether the test is gonna pass

### DIFF
--- a/tests/syscall.rs
+++ b/tests/syscall.rs
@@ -38,7 +38,7 @@ fn syscall() -> CrabResult<()> {
             // Found mapped addr
             test_utils::continue_to_end(&target);
 
-            // unmap the previously mapped memory 
+            // unmap the previously mapped memory
             // and check that it is no longer in the mapped memory list.
             target.munmap(addr as *mut _, len);
             assert!(!target.memory_maps().contains(&addr));
--- a/tests/syscall.rs
+++ b/tests/syscall.rs
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
         )
         .unwrap();
 
+    assert!(target.memory_maps().contains(&addr));
+
     for line in std::fs::read_to_string(format!("/proc/{}/maps", target.pid()))?.lines() {
         if line.starts_with(&format!("{:08x}-", addr)) {
             // Found mapped addr
             test_utils::continue_to_end(&target);
+
+            // unmap the previously mapped memory 
+            // and check that it is no longer in the mapped memory list.
+            target.munmap(addr as *mut _, len);
+            assert!(!target.memory_maps().contains(&addr));
+
             return Ok(());
         }
     }


pro465:
so it seems the tests are not even run, because the only platforms this is being tested in are windows and freebsd (which might be equivalent to linux, idk)

nbaksalyar:
Hmm, let me check this on TravisCI. FreeBSD is quite different and it won't run syscall tests.

bjorn3:
The travis ci config will likely need to be ported to github actions. Travis ci has replaced their free tier with a "trial" plan which they called an "upgrade". (https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) We probably ran out of free credits.

nbaksalyar:
Right, I activated the trial plan for now which should work for ~30 days.
That should give us enough time to move to GH Actions.

I'll trigger a build for this PR manually in a moment.

pro465:
i heard that github actions supports ubuntu builds

nbaksalyar:
OK, this hopefully should trigger a TravisCI build now.

> i heard that github actions supports ubuntu builds

Yeah, the only reason we didn't use it was restricted permissions. To run tests for a debugger, we need to attach to another process using `ptrace` which wasn't available for GH Actions at the time. Situation might've changed so it's worth giving it another try.

pro465:
hmm, that makes sense, thank you!

now, lets see...oh no!

bjorn3:
> Yeah, the only reason we didn't use it was restricted permissions.

Right, forgot about that.

pro465:
I cant even see what specific thing went wrong

tests/syscall.rs
```diff
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
         )
         .unwrap();
 
+    assert!(target.memory_maps().contains(&addr));
```
bjorn3, line 34: Try adding `?` directly after `memory_maps()` here and below.

tests/syscall.rs
```diff
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
...
+
     for line in std::fs::read_to_string(format!("/proc/{}/maps", target.pid()))?.lines() {
         if line.starts_with(&format!("{:08x}-", addr)) {
             // Found mapped addr
             test_utils::continue_to_end(&target);
+
+            // unmap the previously mapped memory
+            // and check that it is no longer in the mapped memory list.
+            target.munmap(addr as *mut _, len);
```
bjorn3, line 43: Also add a `?` right before the semicolon.

tests/syscall.rs
```diff
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
         )
         .unwrap();
 
+    assert!(target.memory_maps()?.contains(&addr));
```
bjorn3, line 34: Maybe something like
```suggestion
    assert!(target.memory_maps()?.iter().any(|map| map.address.0 == addr));
```
works?

tests/syscall.rs
```diff
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
...
+
     for line in std::fs::read_to_string(format!("/proc/{}/maps", target.pid()))?.lines() {
         if line.starts_with(&format!("{:08x}-", addr)) {
```
bjorn3, line 37: By the way it looks like this test was created before `memory_maps()` was introduced. With the assert you introduced this for and if is no longer necessary.

pro465: yes, and for the second part, maybe

pro465:
fyi, i will format it tomorrow, or in other time zones, probably evening.

tests/syscall.rs
```diff
@@ -31,10 +31,18 @@ fn syscall() -> CrabResult<()> {
...
+    assert!(target.memory_maps()?.iter().any(|map| map.address.0 == addr));
+
     for line in std::fs::read_to_string(format!("/proc/{}/maps", target.pid()))?.lines() {
         if line.starts_with(&format!("{:08x}-", addr)) {
             // Found mapped addr
             test_utils::continue_to_end(&target);
+
+            // unmap the previously mapped memory
+            // and check that it is no longer in the mapped memory list.
+            target.munmap(addr as *mut _, len)?;
```
bjorn3, line 43: You can use `1 << 20` as length.

pro465:
so it seems to error because of formatting

pro465:
so it seems that `munmap` is internally panicking...

pro465:
maybe i should remove the libc::MAP_PRIVATE option?

pro465:
@bjorn3 do you know which line of code is trying to access `proc/{pid}/stat` ?

bjorn3:
It is the `Process::new()` call in `fn main_thread()`. I also found the issue. The munmap was placed after `continue_to_end` which causes the process to exit. I pushed a fix. I hope you don't mind.

pro465:
so now it seems it is a completely different (or partially, idk) test that is failing. let me check if i can fix it...

bjorn3:
That is an unrelated test. Maybe travis ci changed a setting?

nbaksalyar (approved):

pro465:
so it seems it really was sporadic...

bjorn3:
This should probably be a squash merge.

pro465:
yeah, with all those formatting and fixing and trying and debugging, i agree with @bjorn3 that this should be squashed into a single commit.

nbaksalyar:
Thanks!

PR #137, opened 2022-04-04
Switch to GitHub Actions for CI

nbaksalyar:
Let's see if the `ptrace` syscall is permitted now.


nbaksalyar:
Seems to be failing because of the `imgui` dependency. I think it'd be reasonable to hide the `gui` example behind a feature flag for now.

bjorn3:
These packages need to be installed for imgui to compile: https://github.com/headcrab-rs/headcrab/blob/ae5e5260f885bf99eba0f035df97d5925f0ff6bb/.travis.yml#L8-L10

nbaksalyar:
Yay, it's working now!

bjorn3 (approved):

PR #139, opened 2022-04-04
Fix race condition in the writemem test

nbaksalyar:
- Update `gimli` and `addr2line` deps.
- Fix race in the `writemem` test: it incorrectly called both `PTRACE_TRACEME` in the child process and `PTRACE_ATTACH` in the parent process while only `PTRACE_TRACEME` is required. This could have resulted in a race condition.
- Expose `LinuxTarget::new` as public API for the cases when we know that the tracee expects to be attached to (i.e. it called `PTRACE_TRACEME`).

### DIFF
--- a/src/target/linux.rs
+++ b/src/target/linux.rs
@@ -184,7 +184,11 @@ impl LinuxTarget {
         Ok(status)
     }
 
-    fn new(pid: Pid) -> Self {
+    /// Creates a new debug target.
+    /// This method should only be used if you know that `pid` is expecting to be traced
+    /// (e.g., when it called `PTRACE_TRACEME`). Otherwise, you should use `LinuxTarget::attach`
+    /// or `LinuxTarget::launch`.
+    pub fn from_debuggee_pid(pid: Pid) -> Self {
         Self {
             pid,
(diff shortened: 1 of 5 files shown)


nbaksalyar:
Closes #116

src/target/linux.rs
```diff
@@ -184,7 +184,11 @@ impl LinuxTarget {
         Ok(status)
     }
 
-    fn new(pid: Pid) -> Self {
+    /// Creates a new debug target.
+    /// This method should only be used if you know that `pid` is expecting to be traced
+    /// (e.g., when it called `PTRACE_TRACEME`). Otherwise, you should use `LinuxTarget::attach`
+    /// or `LinuxTarget::launch`.
+    pub fn new(pid: Pid) -> Self {
```
bjorn3, line 191: Maybe call it new_already_tracing or something like that?

nbaksalyar: Agreed, would be nice to make it more explicit.
How about `LinuxTarget::from_tracee_pid`?

bjorn3: That would work too.

bjorn3 (approved):

